PR #6, opened 2015-12-22
Adds 5.5 Compatibility.

donatj:
For those of us not quite on the bleeding edge yet who still would like tests, this is all that needs to be done to make this work in 5.5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,13 +6,13 @@
 		"preferred-install": "source"
 	},
 	"require": {
-		"php": "^5.6|^7.0",
+		"php": "^5.5|^5.6|^7.0",
 		"jakub-onderka/php-parallel-lint": "~0.8"
 	},
 	"require-dev": {
 		"consistence/coding-standard": "^0.8",
 		"phing/phing": "^2.9",
-		"phpunit/phpunit": "^5.0",
+		"phpunit/phpunit": "^4.0",
 		"satooshi/php-coveralls": "dev-master",
 		"squizlabs/php_codesniffer": "2.5.0 as 2.3.4"
 	},
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: php
 php:
+  - 5.5
   - 5.6
   - 7.0
 before_script:


composer.json
```diff
@@ -6,13 +6,13 @@
 		"preferred-install": "source"
 	},
 	"require": {
-		"php": "^5.6|^7.0",
+		"php": "^5.5|^5.6|^7.0",
```
Vrtak-CZ: :scream: :mask: :question: :bomb:

kukulich:
I really don't think that PHP 5.5 is bleeding edge. PHP 5.6 was released 28 Aug 2014, that's more that 1 year ago. And 5.5 doesn't have active support anymore, see http://php.net/supported-versions.php
So I think it's a bad idea to support 5.5.

donatj:
The 5.5 install base is still very large, much larger than 5.6. - via: [link](http://php.net/supported-versions.php)

screen shot 2015-12-22 at 1 17 23 pm

I think not supporting it even though your tests work fine in it is silly.

kukulich:
From my point of view it's silly to use PHP version that is not actively supported anymore.

We don't use 5.5 so adding support is just pointless work for us.

ondrejmirtes:
I think it's more profitable for everybody to move on to newer versions of PHP and benefit from them. I'd rather spend time improving PHP 7 support (which is the future!) rather than solving compatibility for older versions of PHP. By staying on 5.5, you are doing yourself a disservice and will bump into more and more hurdles as everybody else moves on.

As one can see in your pull request, the compatibility with older versions of PHP is not without drawbacks - you had to downgrade PHPUnit version which means I cannot benefit from goodies in newer versions. Also, I cannot benefit from new features and bugfixes present in PHP 5.6.

If I allowed 5.5 now and wanted to drop it in the future, it would mean cutting off some portion of the userbase and a huge BC break.

My target audience for this (and other tools currently in the making) are serious projects whose developers can influence what version of a language they are on.

ondrejmirtes:
Also, do you know about composer install --ignore-platform-reqs? It's a dangerous workaround with no warranty, but you can try using it for now.

donatj:
I wanted to be able to run some of these in our CI environment. That's way too dangerous.

I'll probably just end up copying the ones I want into our internal code standard. Far from ideal.

Some closing thoughts:

> My target audience for this (and other tools currently in the making) are serious projects whose developers can influence what version of a language they are on.

Those are inversely proportional. The bigger your project, the more inertia in the deployment architecture. Hobbyists can change language versions half a dozen times a day. For large businesses upgrading costs a lot of time, money and man hours. An edge case different that wasn't caught by automated testing can cost tens of thousands of dollars. Jumping versions is not something to be taken lightly.

ondrejmirtes:
Isn't it better to put your efforts to upgrading your infrastructure to PHP
5.6 or at least creating a different 5.6-only CI environment to run this
standard on rather than painfully sticking to 5.5?

janlanger:
@donatj Instead od copying, you can also stick with your fork, keep the PHP 5.5 compatibility there and include it in your composer.json. Since there is currently no 5.6-only code in sniffs, it should be easy to manage (at least for now).

ondrejmirtes:
> For large businesses upgrading costs a lot of time, money and man hours.

Costs of not-upgrading are even higher.

donatj:
How so as long as their are security updates?

ondrejmirtes:
The later you upgrade, the more difficult it is because code bases get bigger over time so there is more (potentially incompatible) code to upgrade.

Also, PHP 7 comes with huge performance gains in CPU performance and memory consumption, so you could even save money on servers.

[See more arguments here](http://blog.ircmaxell.com/2014/12/on-php-version-requirements.html). I'm going to close this PR because I believe it does not make sense to support version of PHP with [already ended active support and security support ending in 6 months](https://secure.php.net/supported-versions.php).

PR #10, opened 2015-12-29
UnusedUsesSniff - false positive

Vrtak-CZ:
false positive return type class without namespace failing test


ondrejmirtes:
Thanks. Look at #11. Please note that in order for UnusedUsesSniff to find uses in annotations, you must set its property `searchAnnotations` to `true`, so the provided failing test would still fail even with the fix.

PR #12, opened 2016-01-06
Normalize array properties

VasekPurchart:
So that configuration options in XML can be written on multiple lines as in `FullyQualifiedClassNameAfterKeywordSniff`, `FullyQualifiedExceptionsSniff` or `ReferenceUsedNamesOnlySniff`.

With this normalization for example namespaces can be configured in a much more readable and diff-friendly way:
```xml
	<rule ref="SlevomatCodingStandard.Files.TypeNameMatchesFileName">
		<properties>
			<property name="rootNamespaces" type="array" value="
				SlevomatCodingStandard => SlevomatCodingStandard,
				tests => SlevomatCodingStandard
			"/>
		</properties>
	</rule>
```

### DIFF
--- a/SlevomatCodingStandard/Helpers/SniffSettingsHelper.php
+++ b/SlevomatCodingStandard/Helpers/SniffSettingsHelper.php
@@ -20,4 +20,23 @@ class SniffSettingsHelper
 		return array_values($settings);
 	}
 
+	/**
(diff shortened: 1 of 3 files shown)


Vrtak-CZ:
:heart:

PR #25, opened 2016-04-15
False positive test for unused uses

Vrtak-CZ:
Add false positive test for unused uses if class in inside annotation


Vrtak-CZ:
Any progress?

ondrejmirtes:
@Vrtak-CZ Good news everyone :) Fixed with https://github.com/slevomat/coding-standard/commit/09976a1683ab31bf99fce9f8821e943a943960c0, merged your test as https://github.com/slevomat/coding-standard/commit/cfe6e2666f2d2c1d3389faf1e2c4770cd1f484be

Issue #27, opened 2016-04-16
Unused use warning for class in type comment


meridius:
As far as I know, there is no support for parsing type comments by _SlevomatCodingStandard.Namespaces.UnusedUses_.

Consider this, which will be found as error _SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse_
```php
use Bar\Something;
use Foo\MyType;

...

/* @var $variable MyType */
$variable = Something::create();
```

Could support for that be added please?

ondrejmirtes:
Do you have searchAnnotations set to true as advised in #22? Can you send failing test or PR?

meridius:
Sure I have searchAnnotations enabled. ;)
Here is the failing test.

PR #28, opened 2016-04-16
Added failing (so far) test for unused uses warning on type comments

meridius:
Referencing https://github.com/slevomat/coding-standard/issues/27

### DIFF
--- a/tests/Sniffs/Namespaces/UnusedUsesSniffTest.php
+++ b/tests/Sniffs/Namespaces/UnusedUsesSniffTest.php
@@ -86,6 +86,14 @@ class UnusedUsesSniffTest extends \SlevomatCodingStandard\Sniffs\TestCase
 		$this->assertNoSniffError($report, 5);
 	}
 
+	public function testTypeComment()
+	{
+		$report = $this->checkFile(__DIR__ . '/data/unusedUsesTypeComment.php', [
+			'searchAnnotations' => true,
+		]);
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
@meridius What about using `/**` as the beginning of the comment? It's a different token for PHPCS and the defacto standard for docblocks.

meridius:
@ondrejmirtes But then it would be docblock, not type comment and I'd get error about docblocks not being able to be only on one line. 
Since I would not be able to test it untill evening, I don't know what sniff exactly reports this, but I agree with it.

ondrejmirtes:
Documentation annotations like `@var` or `@see` are always in a docblock `/** */`, that's why I scan them only. Please verify that the sniff works in that scenario.

meridius:
Well, not always they don't. Please see the Netbeans' documentation article about this feature.

https://blogs.oracle.com/netbeansphp/entry/defining_a_variable_type_in

ondrejmirtes:
It's a design decision on our side to support only one type of notation. And that's `/**` in order to be consistent with property and method docblocks.

See for example https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md - PSR-5 proposal about phpDocs - it also supports `/**` only.

Using `/**` can benefit you also when you are using Doctrine annotations, which parse `/**` only too.

I'm going to close this issue for now - if this sniff does not see types inside `/**` or if you have other problems, please open a new one. Thanks.

PR #34, opened 2016-04-22
Disallow use from root namespace

kukulich:


kiding:
Can this very change be enforced using PHP_CodeSniffer, or any other tool available? #245 suggests Slevomat Coding Standard can not, but I was wondering if @kukulich has a suggestion.

mssimi:
@kiding https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/rules/import/global_namespace_import.rst

PR #43, opened 2016-06-14
Failure test

jirastom:
Case of reference name join and use statement Join do not match (SlevomatCodingStandard.Namespaces.UnusedUses.MismatchingCaseSensitivity)


ondrejmirtes:
Thanks! Fixed with https://github.com/slevomat/coding-standard/commit/0a12b4a1fda0157d199a1e0e8f8c163902eba573 and merged as https://github.com/slevomat/coding-standard/commit/58dafeac2598a8d98a81c55d28c234f798588ba1

PR #45, opened 2016-07-11
False positive for private property used in string

voda:
Hey, found a bug where private properties are marked as unused even if they used in double quoted strings `"{$this->foo}"`.


ondrejmirtes:
This is not trivial to fix. Feel free to send a PR that would work with all the ways that properties can be accessed inside strings. So `{$this->foo . $this->bar}` also works etc.

Or use a different way how to concatenate strings dynamically. Our preferred way to do this is with `sprintf`.

PR #46, opened 2016-07-21
Probably bugs

ondrejmirtes:


ondrejmirtes:
`public function stringParameterWithoutHint($cacheKey)` is not a bug because PHP_CodeSniffer does not have access to reflection, only to tokens of the current file.

PR #50, opened 2016-08-06
composer: autoload sniffs so they can be used in 3rd party apps

TomasVotruba:
This is needed, if I want to use those sniffs via composer autoload, and not custom PHP_CodeSniffer autoload.

Ref: https://github.com/Symplify/PHP7_CodeSniffer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,9 +17,14 @@
 		"phpunit/phpunit": "^5.0",
 		"satooshi/php-coveralls": "^1.0"
 	},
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thanks 👍  Using these sniffs as dependencies in other way than via include `ruleset.xml` didn't occur to me, hope they will be useful.

TomasVotruba:
Thank you! Helps a lot

PR #54, opened 2016-08-09
Fix false positive warning about missing type hint by defining extern…

pepakriz:
…al namespaces


ondrejmirtes:
@pepakriz I don't like it because there's no way to deal with multiple-level inheritance hierarchy:

Foo <-- Bar <-- External\BaseClass

Missing typehints in `Foo` would still be reported.

There's an annotation that can be used above the method:

```
@phpcsSuppress SlevomatCodingStandard.Typehints.TypeHintDeclaration.missingParameterTypeHint
```

PR #63, opened 2016-10-25
Better Classes.UnusedPrivateElements Sniff

o5:
Hello, 

I think, sniff `SlevomatCodingStandard.Classes.UnusedPrivateElements` should not warn, if I use private property in string. I don't remember name of general rule which it prohibits, but I think some new option like `alwaysUsedPropertiesAnnotations` would be great. Hmm? :)

Thanks

### DIFF
--- a/tests/Sniffs/Classes/data/ClassWithSomeUnusedProperties.php
+++ b/tests/Sniffs/Classes/data/ClassWithSomeUnusedProperties.php
@@ -9,6 +9,8 @@ class ClassWithSomeUnusedProperties extends \Consistence\Object
 
 	private $usedProperty;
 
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
I think it's very difficult to come up with a solution that would cover all possible ways how to access a class property in PHP to make this sniff bulletproof. I don't want to write a PHP parser :)

Instead, I would recommend you to use sprintf() in these cases. If you want to continue using {}, you will be able to use phpcsSuppress annotation in upcoming 2.0 version thanks to [this commit](https://github.com/slevomat/coding-standard/commit/39dc6e3102130ce41c0e0f3b6dd47c89a556dd7f).

PR #64, opened 2016-10-26
EmptyLinesAroundTypeBrances: added tests, change to fixable error

f3l1x:
PR for #49


ondrejmirtes:
I think that the fixer does not work very well in all situations - it will fix when there's a newline missing, but not when there are more lines than expected. Can you fix that? Thanks.

f3l1x:
@ondrejmirtes Thank you for the feedback. I'll try to.

f3l1x:
@ondrejmirtes I'm quite busy these days. If you would like to resolve it, do not hesitate.

kukulich:
@f3l1x Thanks, I've used some of your code in https://github.com/slevomat/coding-standard/pull/79.

PR #67, opened 2016-11-04
UnusedUsesSniff - fixed multiple exceptions in catch

JanKonas:
Updated `ReferencedNameHelper` to recognize all exception classes within php 7.1 [multiple exception catch](https://wiki.php.net/rfc/multiple-catch).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: php
 php:
   - 5.6
   - 7.0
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
@JanKonas Awesome job, thank you! I didn't even know about the >= 7.1 annotations for lint and PHPUnit :)

PR #68, opened 2016-11-14
Support for multiple extensions

klapuch:
In the current release it is not possible to run phpcs against Nette Tester tests without fail, because the tester requires files with .phpt extension. So, every time I try to run phpcs I end up with fail as namespace does not match with filepath.


kukulich:
I've cherry picked your commit: https://github.com/slevomat/coding-standard/pull/98

PR #72, opened 2016-12-23
[2.0] AnnotationHelper does not handle parametrized annotations properly [failing-test]

mhujer:
I'm getting invalid `SlevomatCodingStandard.Typehints.TypeHintDeclaration.uselessDocComment` error.

With the following setup complex annotation (such as `@Route("/")` are not parsed properly and therefore the docblock with such annotation is seen as useless.
```
    <rule ref="SlevomatCodingStandard.Typehints.TypeHintDeclaration">
        <properties>
            <property name="usefulAnnotations" type="array" value="
                @Route
            "/>
        </properties>
    </rule>
```

This PR contains failing test (passes when there is space after `@Route` in `annotation.php`)

### DIFF
--- a/tests/Helpers/AnnotationHelperTest.php
+++ b/tests/Helpers/AnnotationHelperTest.php
@@ -52,6 +52,13 @@ class AnnotationHelperTest extends \SlevomatCodingStandard\Helpers\TestCase
 		$this->assertSame('string $b', $annotations[1]);
 	}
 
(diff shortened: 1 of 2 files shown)


kukulich:
Thanks. Fixed in https://github.com/slevomat/coding-standard/commit/86678f74eb8a64fc450a9af07f7c35397084b0c2

mhujer:
Great, thanks! 👍

PR #73, opened 2016-12-23
[2.0] DeclareStrictTypesSniff fix for empty file

mhujer:
When the file contains only opening tag the sniff fails with *undefined index*


mhujer:
Build error does not seem related to this PR

kukulich:
Merged, thanks.

PR #87, opened 2017-02-02
Incorrect detection of Yoda in expression beginning with cast

VasekPurchart:
If there is a cast on the left side of the expression, this is incorrectly detected as a yoda comparison:

```php
<?php

(int) $foo === $bar;
```
results in:
```
 3 | ERROR | [x] Yoda comparisons are prohibited.
```

I have tried to fix this but I am not sure why there was a dynamism value of `3`, which is the thing I think throws this off balance. I have tried to replace this with the "dynamic" dynamism, which fixes my issue, but I am not sure about other cases, see:

```php
<?php

(int) FOO === $bar;
```

I consider this to be a Yoda statement, but with that dynamism change it would not be  - that is why the tests fail now.

When I thought about this I think the cast should be "skipped" for the purpose of the evaluation, because they don't change the nature of the expressions regarding its dynamism? What is your opinion?

cc @ondrejmirtes

### DIFF
--- a/SlevomatCodingStandard/Sniffs/ControlStructures/YodaComparisonSniff.php
+++ b/SlevomatCodingStandard/Sniffs/ControlStructures/YodaComparisonSniff.php
@@ -20,6 +20,8 @@ class YodaComparisonSniff implements \PHP_CodeSniffer_Sniff
 
 	const DYNAMISM_FUNCTION_CALL = self::DYNAMISM_VARIABLE;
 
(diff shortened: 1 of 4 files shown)


kukulich:
Fixed in 2.0-dev in this commit: https://github.com/slevomat/coding-standard/pull/103/commits/d8fac78ed9b611eaaaaa043af714a333009fe0fb

PR #105, opened 2017-02-13
Added NullableTypeForNullDefaultValueSniff

janedbal:
- Enforces using nullability symbol `?` when using `null` as default value, e.g. `bool $param = null` > `?bool $param = null`. 
- I'm not sure with the naming, any better ideas?
- Reason to enforce it: you should use nullability symbol if parameter accepts null value no matter what the default value is. So, if you change from `?bool $param = false` to `?bool $param = null`, you don't need to change typehint - you are changing only default value.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/TypeHints/NullableTypeForNullDefaultValueSniff.php
+++ b/SlevomatCodingStandard/Sniffs/TypeHints/NullableTypeForNullDefaultValueSniff.php
@@ -1,0 +1,82 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\TypeHints;
+
+use SlevomatCodingStandard\Helpers\TokenHelper;
+
+class NullableTypeForNullDefaultValueSniff implements \PHP_CodeSniffer_Sniff
+{
+
+	const CODE_NULLABILITY_SYMBOL_REQUIRED = 'NullabilitySymbolRequired';
+
+	/**
+	 * @return int[]
+	 */
+	public function register(): array
+	{
+		return [
+			T_FUNCTION,
+			T_CLOSURE,
+		];
+	}
+
+	/**
+	 * @phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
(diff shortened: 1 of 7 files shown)


VasekPurchart:
What about `NullableTypeForNullDefaultValue`? Since this (nullable type) is the name of the feature as per the [RFC](https://wiki.php.net/rfc/nullable_types) (and I have not found any mentions in the docs).

SlevomatCodingStandard/Sniffs/TypeHints/NullabilityWhenNullDefaultValueSniff.php
```diff
@@ -0,0 +1,79 @@
...
+
+				$ignoreToken1 = array_merge(TokenHelper::$ineffectiveTokenCodes, [T_BITWISE_AND, T_ELLIPSIS]);
+				$typehint = TokenHelper::findPreviousExcluding($phpcsFile, $ignoreToken1, $i - 1, $startPointer);
+
+				if ($typehint === null || !in_array($tokens[$typehint]['code'], [T_NS_SEPARATOR, T_STRING], true)) {
+					continue;
+				};
+
+				$ignoreTokens2 = array_merge(TokenHelper::$ineffectiveTokenCodes, [T_BITWISE_AND, T_ELLIPSIS, T_NS_SEPARATOR, T_STRING]);
+				$beforeTypehint = TokenHelper::findPreviousExcluding($phpcsFile, $ignoreTokens2, $i - 1, $startPointer);
```
kukulich: The third parametr should be `$typehint - 1`. Then you don't need to ignore `T_BITWISE_AND` and `T_ELLIPSIS` here and you can use the same list of type hint tokens from line 51.

kukulich:
@janedbal You can squash your commits. It think one commit should be enough here.

janedbal:
@kukulich Fixed, thank you! 

I'll wait with the squash for your approval so you can validate the changes more easily (divided in commits). Or you can use squash on GitHub :)

kukulich:
@janedbal Superb! You can squash it now. Please don't forget to change commit message so there's the new sniff name :)

janedbal:
@kukulich Squashed

PR #107, opened 2017-02-21
[failing test only] complex annotation can be marked as useful

mhujer:
Complex annotations such as `@Assert\Callback()` cannot be marked as useful, see the failing test.

**Workaround:** Add description to the docblock such as _Remove this text when the `@Assert\Callback()` annotation can be marked as useful_


kukulich:
@mhujer If you add `@Assert\Callback` to your ruleset.xml, it should work. However I may add support for prefixes...

mhujer:
It fails as well - I've added another commit

kukulich:
Ok, I'll fix it in the evening :)

kukulich:
@mhujer It should be fixed. There's also prefix support now so you can use `Assert\` in your ruleset.xml.

mhujer:
@kukulich It works great, thank you very much! 👍

PR #110, opened 2017-03-06
[failing test] Class used as targetEntity should not be reported as unused

mhujer:
When the class is used as `targetEntity` in Doctrine annotation, it is reported as unused and it should not be. (PHPStorm detects this properly)


kukulich:
@mhujer Can you please try this branch? https://github.com/slevomat/coding-standard/pull/112

mhujer:
@kukulich It works fine with this branch 👍

kukulich:
Fixed in https://github.com/slevomat/coding-standard/pull/112.

PR #111, opened 2017-03-06
Fix of bug in allowFullyQualifiedNameForCollidingClasses flag for ReferenceUsedNamesOnlySniff

jost125:
There were bug in finding referenced classes in their index. Instead of looking into keys I have looked into values, which of course cannot work.


kukulich:
Thanks.

PR #114, opened 2017-03-11
Detect property reads inside double quoted strings

jost125:
Previous version of sniff was not detecting property reads inside of double quoted strings (unfortuntelly, phpcs internal tokenizer does not tokenize content of double quoted strings - manual parsing needed to be done).


kukulich:
@jost125 I think it should be easier to use `preg_match_all` to string parsing or did you find a reason why to use `token_get_all`?

jost125:
I can use preg_match_all, but I am not sure, if that would be more readable. Somethig like this might work with negative lookaheads and negative lookbehinds https://3v4l.org/YETtg

BTW I just noticed that I am not looking for following brackets after string (I will send fixup).

kukulich:
@jost125 What you think https://github.com/slevomat/coding-standard/commit/c2ef414478005d1711a49d459495e2c571757616 ?

jost125:
Great! :+1:

kukulich:
Fixed in https://github.com/slevomat/coding-standard/commit/46ae8aa75f773116c5718b6475a20e7ee00a665a

jost125:
Thanks.

PR #124, opened 2017-04-06
composer: drop minimal stability

TomasVotruba:
Since all used packages are stable, this is not really needed. Is it?


kukulich:
@TomasVotruba Do you know any situation when actual version can cause a problem?

TomasVotruba:
It's not issue fix. It's redundant code removal.

What's the reason for it?

kukulich:
I don't have to deal with it when I test newer versions of dependencies.

TomasVotruba:
I see, it makes sense.

PR #129, opened 2017-04-21
ReturnTypeHintSpacingSniff: Allow configurable spacing before colon

Majkl578:
This PR adds an option to specify number of spaces before return type colon, using the `$spacesCountBeforeColon` property.
Defaults to 0 which is the current behavior and thus should be fully backward compatible.

Typical use case is code like this:
```php
function foo() : int
{
}
```


kukulich:
@Majkl578 Very nice! Can you please only use variable for the str_repeat code?

Majkl578:
Done. Hopefully also fixed integer normalization...

PR #130, opened 2017-04-21
TypeHintDeclarationSniff: Respect inheritance when suggesting adding native type hints (failing test only)

Majkl578:
The TypeHintDeclaration.MissingParameterTypeHint currently does not verify whether the typehint could be added or not.
When the ancestor has a parameter without type (i.e. Symfony, Doctrine and other PHP5-compatible code), it still wants to add the type, even though it'd be a violation of the inheritance.

Only failing test added (at least for now).


ondrejmirtes:
This is not possible to do in PHPCS, that's why there's a suppress annotation available. See https://medium.com/@ondrejmirtes/three-pillars-of-static-analysis-in-php-f3f5d7bfd61b

kukulich:
Yes, it's unfortunately not possible.

PR #131, opened 2017-04-21
EmptyLinesAroundTypeBracesSniff: Allow configurable number of lines around braces

Majkl578:
This PR allows configurable number of empty lines after opening brace and closing brace, using `$linesCountAfterOpeningBrace` and `$linesCountAfterOpeningBrace` respectively.

Both default to 1 empty line which is current behavior, thus BC should be preserved.

Motivation is simple, there is a code out there that requires either no or 2-3 spaces.


kukulich:
@Majkl578 Perfect! Thank you very much.

PR #132, opened 2017-04-21
ReferenceUsedNamesOnlySniff: Add option to allow fully-qualified global classes

### LINKED ISSUE #76: Disable ReferenceViaFullyQualifiedName for certain types
Type \DateTime should not be referenced via a fully qualified name, but via a use statement (SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName)

Is there a way to disable it for this type? Thanks!

Majkl578:
This should resolve #76. Setting `allowFullyQualifiedGlobalClasses` to `true` should enable ignoring FQCN from global namespace.

Many people (myself included) prefer FQCN for classes from global namespace rather than putting them in use.

The change seems a bit too simple for such complex code so please check whether it's enough or something else needs changing. 😊


kukulich:
@Majkl578 Superb. I haven't time to do it myself so thank you very much.

Issue #138, opened 2017-04-24
False positive on UnusedUsesSniff


tremby:
I have a `use` statement for `GuzzleHttp\Promise`. In my code I'm using a bare function from that namespace, as `Promise\settle()`. The `UnusedUsesSniff` is complaining that I'm not using `GuzzleHttp\Promise`, which I believe to be false.

I also tried `use GuzzleHttp\Promise\settle` and then using `settle()`, and I get an error from the sniff here too.

kukulich:
@tremby Could you please post here your ruleset.xml?

tremby:
I don't understand why, but OK.

```xml
<?xml version="1.0"?>
<ruleset name="Tribal.Laravel5">
  <description>
    Ruleset for PHP projects using Laravel 5
  </description>

  <!-- Referencing PSR2 includes the whole PSR2 standard -->
  <rule ref="PSR2"/>

  <rule ref="Generic.Arrays.DisallowLongArraySyntax"/>
  <rule ref="vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/Arrays/TrailingArrayCommaSniff.php"/>
  <rule ref="vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/Namespaces/AlphabeticallySortedUsesSniff.php"/>
  <rule ref="vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php"/>
</ruleset>
```

Majkl578:
True, looks like it doesn't work with functions in combination with partial uses.
Simple repro code:
```php
<?php

use GuzzleHttp\Promise;

Promise\settle($promise);
```
produces:
```
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 3 | ERROR | [x] Type GuzzleHttp\Promise is not used in this file.
   |       |     (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

Majkl578:
@tremby: Could you test with  #139?

tremby:
@Majkl578: works for me, and verified that the sniff still complains when it should (at least in trivial cases). Looking forward to a merge and tag.

PR #139, opened 2017-04-24
UnusedUsesSniff: Fix detection of partially referenced functions and constants

Majkl578:
Fix for #138.


kukulich:
@Majkl578 Great work again! However I'm too drunk now so I will review it rather tomorrow again :)

PR #143, opened 2017-04-26
Move build related configurations to build dir

VasekPurchart:
* `ruleset.xml` in root can be confusing, because it is not the standard, that this repository is "providing"
* cleans up the root dir


Majkl578:
> ruleset.xml in root can be confusing

I agree it's sort of confusing, why not using phpcs.xml?

VasekPurchart:
We can, but the main reason for this is the second reason and also #144 :)

PR #145, opened 2017-04-26
TestCase: Obtain Sniff path from reflection instead of hardcoded path

Majkl578:
Since my sniffs are heavily based on your approach, I also took the opportunity to use your TestCase with an easy-to-use assertion logic.
The only issue was that I had to overrride TestCase::getSniffPath() to work in a more generic way.

This PR keeps the behavior, but allows consumers to smoothly use your TestCase for their own sniffs, without the need to overrride the method.

### DIFF
--- a/tests/Sniffs/TestCase.php
+++ b/tests/Sniffs/TestCase.php
@@ -30,11 +30,9 @@ abstract class TestCase extends \PHPUnit\Framework\TestCase
 		$codeSniffer->registerSniffs([$this->getSniffPath()], [], []);
 
 		if (count($codesToCheck) > 0) {
-			$classReflection = new \ReflectionClass($this->getSniffClassName());
-
 			$ruleset = $propertyReflection->getValue($codeSniffer);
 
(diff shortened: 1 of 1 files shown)


tests/Sniffs/TestCase.php
```diff
@@ -157,14 +157,7 @@ protected function getSniffClassName(): string
 
 	protected function getSniffPath(): string
```
kukulich: Can you please remove this method and create the reflection right in method `checkFile`? The same reflection is on line 33 so it would be nice to create it just once.

Majkl578: That could break existing code though...

kukulich: @Majkl578 You're right. So new method `getSniffClassReflection()` with `static $reflectionClass` inside?

Majkl578: Done.
Made it private at first and was wondering why it doesn't work... 🥇

PR #148, opened 2017-04-27
ClassConstantVisibilitySniff: Make the sniff fixable

Majkl578:
It's safe to assume the constant should have public visibility, because it's the default. ☺️ 

Granted, not all previously public constants should stay public, but it's up to the developer to decide (later).


Majkl578:
Hmm, how do I make it work under 7.0? The `fixed` file is being linted and obviously fails... :/

ondrejmirtes:
I'm not sure about this. When we transitioned to PHP 7.1, we had to decide
with each constant whether to make it public or private. This fixer doesn't
take any advantage of the new language feature, it avoids it...

kukulich:
I had this idea earlier too but then I decided it's better not to implement this. Everyone should decide with each constant whether to make it public, protected or private. And if someone wants to make all constants public and decide about them later, it's very easy to do it with search&replace functionality in every editor. So I don't want to support this "bad" fix because there's another easy way how to "fix" it :)

Thank for you effort but this PR is not going to be merged.

TomasVotruba:
I think making it default `public` make sense.

When use this sniff, I probably want to add this feature. I'd expect help from coding standard. Using `public` is help that doesn't break anything and let me decide manually, where I want to use `public` or `protected`.

Using `protected` or `private` by default might break app, so I would not use them.

Already made in Fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2213

PR #150, opened 2017-04-30
Added InlineDocCommentDeclarationSniff

kukulich:


kukulich:
cc @VasekPurchart

VasekPurchart:
```
 100 | ERROR | [x] Invalid inline doc comment format "@var
     |       |     $annotationDriver
     |       |     \Doctrine\ORM\Mapping\Driver\AnnotationDriver" for
     |       |     variable $annotationDriver.
```

Works great, bud maybe let's add to the message, what is the correct form? So something like `Type should precede the variable name`?

kukulich:
I've just improved the error message.

VasekPurchart:
👍

PR #153, opened 2017-05-04
Add option to sort uses case sensitively

VasekPurchart:
With this setting the namespaces are compared "by words", grouping words together similarly as namespaces.

I've also fixed the outdated example in the readme and replaced it with more natural examples, which I believe better show the difference one might encounter.


Majkl578:
This makes more sense, but is not compatible with PhpStorm unfortunately. :/

VasekPurchart:
Yes, we have been discussing that, but from what we found out:

1) The ocasions where you actually encounter this are pretty rare (a few single cases in huge Slevomat codebase)
2) There is no documented way in PHPStorm as to wheter this is the intentional (someone thought it trough) sorting or just an implementation detail.

Since this is more natural (when you type it manually) and in line Consistence, I'll be using that there. And what you know, maybe PHPStorm could change the algo as well someday :)

PR #156, opened 2017-05-12
Splits CODE_MISSING_PARAMETER_TYPE_HINT into two

OndrejVasicek:
For some reason constant `CODE_MISSING_PARAMETER_TYPE_HINT` is used for two different situations – one is for missing the doc and the hint in the same time, the other one is for suggesting using hint, because it should be possible (the message says).

In many cases it’s OK and I should always use hint, but in my case I just inherited a method from a parent, who isn’t mine but it’s from a framework. In this case I’m not able to use hint, because it breaks the PHP rule about the “parameters in all inherited methods must be the same type”. Funny is that setting the return type declarations isn’t a problem. 

So in this PR I’ve split this constant into two separated and thanks to that I’m able to use `exclude ` in my own ruleset and still keep using the annotation for missing hint and docs in the same time.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/TypeHints/TypeHintDeclarationSniff.php
+++ b/SlevomatCodingStandard/Sniffs/TypeHints/TypeHintDeclarationSniff.php
@@ -20,6 +20,8 @@ class TypeHintDeclarationSniff implements \PHP_CodeSniffer_Sniff
 	const NAME = 'SlevomatCodingStandard.TypeHints.TypeHintDeclaration';
 
 	const CODE_MISSING_PARAMETER_TYPE_HINT = 'MissingParameterTypeHint';
+	
+	const CODE_MISSING_PARAMETER_TYPE_HINT_AND_DOC = 'MissingParameterTypeHintAndDoc';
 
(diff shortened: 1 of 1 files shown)


Majkl578:
> I just inherited a method from a parent, who isn’t mine but it’s from a framework.

True. This is a known limitation of PHPCS. It can only analyse a single file at same moment so it can't make any assumptions about inheritance (besides seeing _extends_).

> Funny is that setting the return type declarations isn’t a problem.

It's not actually. Return types are invariant, but sub-types could add the return type as long as it's not specified in parent. This does not break LSP rules.

Note that this change is BC breaking as it will break previous suppressions done by:
```
@phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
```

VasekPurchart:
As @Majkl578 stated plus you would be better off not ignoring whole rule due to one exception, rather than that use a local suppress (as in the example Majkl posted).

kukulich:
Thanks @Majkl578 and @VasekPurchart for explanation why this PR is not going to be merged.

OndrejVasicek:
Thanks guys for the explanation.

PR #176, opened 2017-08-07
False positive for UnusedMethod when self::privateMethod() is used

pepakriz:
I discovered another one false positive for static method. Failing test is attached.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/a20058d08da39925177a67ce908cb381647c82d4

PR #205, opened 2017-10-16
add failing test useless docblock

bendavies:
Hey,

There's a bug in detecting useless docblock when parameters are aligned, as in my failing test case.
I did try to fix it, but ran out of time trying to fix your regex.

Hopefully you can fix it faster than me :)

Cheers.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/0dc2a93eed4b0bc4c76e3e2f05f2d1d86d76042b

PR #216, opened 2017-11-20
Add failing test case for trait method renaming

alcaeus:
When renaming a method while importing a trait, the `FullyQualifiedGlobalConstantsSniff` class incorrectly reports the two method names as global constants, which results in errors and potentially failing code when running `phpcbf`.

I don't have an idea how to fix this but thought I'd provide a failing test case to get started.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/b95f6344c43c0da142bb5cb97fcba6310f062b0a.

alcaeus:
Thank you! ❤️

PR #231, opened 2017-12-30
Move base TestCase to src/ so consumers can use it

Majkl578:
Moving base TestCase with helpers from tests/ to src/ In order to allow its use in dependent code. Currently the only option is `prefer-source` for installing this package, since tests/ is `export-ignore`d.


Majkl578:
Huh? Coverage decrease looks like false positive?

kukulich:
Cherry picked :)

Majkl578:
Thanks!

PR #233, opened 2018-01-03
FullyQualifiedGlobal{Constants,Functions}Sniff: Refactor logic into shared ancestor

Majkl578:
This was originally part of my larger PR to support global function/constant uses through `use` and expanding it by forbidding either `use` or `\` + fixability - in order to force all global functions be references in `use function`. I haven't finished it yet though, making such cases auto-fixable is pretty hard. So I'll likely drop the fixability part of those changes for now and only submit the configuration part.

I hope I didn't mess up rebase after f6667fac2306fed1fad6d0aee4f7f32ecc863ddc + 8af6030583afa590fcbcbcd7f52b36376a46186f.


kukulich:
Thanks.

PR #235, opened 2018-01-03
FullyQualifiedGlobal{Constants,Functions}Sniff: Configurable

Majkl578:
Allow imports and fully qualified references to be configured.

Use case:
* enforce only use-only references:
```php
use function Foo\bar;

bar(); // ok
\Foo\bar(); // not ok
```
* enforce only fully-qualified-only references:
```php
use function Foo\bar;

bar(); // not ok
\Foo\bar(); // ok
```

Default behavior still allows both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,8 @@ All references to global constants must be referenced via a fully qualified name
 Sniff provides the following settings:
 
 * `exclude`: list of global constants that are allowed not to be referenced via FQN.
+* `allowImported`: whether to allow `use const`
+* `allowFullyQualified`: whether to allow in-place fully qualified references (i.e. `\Foo\BAR`)
 
 #### SlevomatCodingStandard.Namespaces.FullyQualifiedGlobalFunctions 🔧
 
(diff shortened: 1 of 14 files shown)


Majkl578:
Note: Originally wanted to make these fixable, but there are too many edge cases and it became too bloated. So for now it's report-only. Feel free to suggest a way.

kukulich:
The original sniffs were intended only to check global functions and constants. So `use function Foo\bar` is not ok because it's not global function.

These sniffs even have "global" in their names so I don't think it would be good to add this functionality to them. I think it's better to add new sniffs, each with its own small purpose. We already have some big sniffs with a lot of options and they are really complicated to understand.

Majkl578:
I see your point, but this behavior of namespaced functions/constants through `use` is allowed even without this PR: https://github.com/slevomat/coding-standard/blob/master/tests/Sniffs/Namespaces/data/fullyQualifiedGlobalFunctionsNoErrors.php#L6
This PR does not touch that behavior afaik, it only adds option to forbid either `use` or `\`.
So what do you suggest?

kukulich:
You've misunderstood the test. The code `use function Used\hihi;` is there only to be sure that this line is not reported as error or changed during fixation. The sniff doesn't allow or disallow this, the sniff simply ignores it :)

Majkl578:
I give up, completely broken by conflicts from 5b806df4ab28dc26752a213c0302b111ee4a188f.

janedbal:
I was really looking forward for this feature. IMHO it could be solved by better naming of the sniff. The problem is that the original sniff was implemented for too narrow purpose and Majkl's update therefore doesn't completely fit in. What do you think?

kukulich:
@janedbal Please move the discussion to https://github.com/slevomat/coding-standard/issues/236

PR #256, opened 2018-02-01
DocCommentHelper: Trim description, test anonymous/inline descriptions

Majkl578:
Just a little addition to 170deb0bcc78f89dc3c1ccb3aa61837cc45a3fb3.

In case of the inline comment, there is a space at the end - it may also be a PHPCS bug?

### DIFF
--- a/tests/Helpers/DocCommentHelperTest.php
+++ b/tests/Helpers/DocCommentHelperTest.php
@@ -132,6 +132,22 @@ class DocCommentHelperTest extends \SlevomatCodingStandard\Helpers\TestCase
 		);
 	}
 
+	public function testUnboundGetDocCommentDescription(): void
+	{
+		self::assertSame(
+			['Created by Slevomat.'],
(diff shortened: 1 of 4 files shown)


tests/Helpers/DocCommentHelperTest.php
```diff
@@ -132,6 +132,22 @@ public function testFunctionGetDocCommentDescription(): void
...
 
+	public function testunboundGetDocCommentDescription(): void
+	{
+		self::assertSame(
+			['Created by Slevomat.'],
+			$this->stringifyComments(DocCommentHelper::getDocCommentDescription($this->getTestedCodeSnifferFile(), $this->findPointerByLineAndType($this->getTestedCodeSnifferFile(), 3, T_DOC_COMMENT_OPEN_TAG)))
+		);
+	}
+
+	public function testunboundMultilineGetDocCommentDescription(): void
```
kukulich: Typo, twice :)

Majkl578: Ugh, need a sniff for that. 😅

PR #259, opened 2018-02-04
Private methods used in heredoc are not recognized (failing test)

alcaeus:
`SlevomatCodingStandard.Classes.UnusedPrivateElements` does not recognize private methods used within HEREDOC strings. This possibly extends to properties and other private members as well.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/61587d8d9c2301d83ba967bb596f279e58b4bf1e

alcaeus:
Thanks for the quick fix!

PR #262, opened 2018-02-06
add failing test for early exit - condition cannot be moved any further

pilec:
If I have condition inside of foreach loop, I don't know the actual value of property, so I cannot to rewrite this piece of code to be able to pass phpcs check.

### DIFF
--- a/tests/Sniffs/ControlStructures/data/earlyExitNoErrors.php
+++ b/tests/Sniffs/ControlStructures/data/earlyExitNoErrors.php
@@ -133,3 +133,12 @@ for ($i = 0; $i < 100; $i++) {
 
 	doSomethingAgain();
 }
+
+function passWhenConditionIsOnTheEndOfMethod()
+{
+	for ($i = 0; $i < 100; $i++) {
+		if ($i % 2 === 0) {
+			doSomething();
+		}
+	}
+}


pilec:
Thanks 👍 , you are completely right - little bit harder to read still, but ok.

kukulich:
@pilec You can always disable the sniff for some piece of code :)

Eg. https://github.com/Roave/BetterReflection/blob/master/src/NodeCompiler/CompileNodeToValue.php#L290

paveljanda:
@kukulich Your code is really much more complex to read and understand. :)

Compare following sentences:

- "Go eat something if you're hunry"
- "If you are not hungry, do not eat. But if different situation occurs, go eat something."

You see the difference? Just saying. :)

pilec:
@kukulich I know, but I would like to be as strict as possible. Btw I met another example:

This one produce an error
```
foreach (['foo', 'bar'] as $identity) {

	if ($data === 'foo') {
		break;
	}
}
```

This one not:
```
```
foreach (['foo', 'bar'] as $identity) {

	if ($data === 'foo') {
		break;
	}

	continue;
}
```

kukulich:
@pilec Yes, it really looks like an error :) Thanks, I will fix it.

kukulich:
Fixed in https://github.com/slevomat/coding-standard/commit/fb29dad05098fa549b911fc13de3d202347c49a1

pilec:
Works like a charm! Thanks 👍

PR #271, opened 2018-02-07
Fix UseStatementHelper::isTraitUse for anonymous function inside method

grongor:


kukulich:
@grongor Did I understand correctly that the second commit is fix for the first commit? If so, please squash it to just one commit :) Thanks.

grongor:
Yeah it is, I wanted to show that it failed before :) I'll also fix missing `;` :D

kukulich:
Thanks.

Issue #278, opened 2018-02-08
UnusedUsesSniff and case mismatch


gsdevme:
Hi,

I can see via https://github.com/slevomat/coding-standard/commit/7795645f18aebbff58ae1d0062ec6590642d0b3f there was a change to search for case-mismatch.. however Im not sure if this is a behaviour change. In the example below its flagged because `exception` !== `Exception` however.. this is a just a documentation comment about the class.. Was this meant to target types in docblocks only?

Documentation like this tends to be written in non-programming syntax, whats the thought behind this change long term?

```php

use Exception;
use Symfony\Component\HttpFoundation\Response;

/**
* An exception that encodes a fully-formed HTTP response.
*/
class ResponseException extends Exception
```

Another example, a simple documentation comment written in english.

```php
/**
 * A custom Doctrine type for personal details in a checkout.
 */
class PersonalDetailsType extends Type
```

Majkl578:
I can confirm this issue, lots of false positives. For example [for Doctrine's CacheFactory](https://github.com/doctrine/doctrine2/blob/4509c770da4dbed1568ddd3f89bd0663c01ef059/lib/Doctrine/ORM/Cache/CacheFactory.php):
```
 23 | ERROR | Case of reference name cache and use statement Cache do not match.
 58 | ERROR | Case of reference name cache and use statement Cache do not match.
 58 | ERROR | Case of reference name region and use statement Region do not match.
 61 | ERROR | Case of reference name region and use statement Region do not match.
 63 | ERROR | Case of reference name cache and use statement Cache do not match.
 88 | ERROR | Case of reference name cache and use statement Cache do not match.
 88 | ERROR | Case of reference name region and use statement Region do not match.
 90 | ERROR | Case of reference name cache and use statement Cache do not match.
 92 | ERROR | Case of reference name cache and use statement Cache do not match.
 97 | ERROR | Case of reference name cache and use statement Cache do not match.
 97 | ERROR | Case of reference name region and use statement Region do not match.
 99 | ERROR | Case of reference name region and use statement Region do not match.
```

Soullivaneuh:
Same thing here:

```
FILE: /code/src/Event/AbstractHookEvent.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
-----------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #282, opened 2018-02-11
UnusedUsesSniff: add option to ignore specified annotation names

xificurk:
For more details about the motivations see discussion in #278

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
@@ -19,6 +19,9 @@ class UnusedUsesSniff implements \PHP_CodeSniffer\Sniffs\Sniff
 	/** @var bool */
 	public $searchAnnotations = false;
 
+	/** @var string[] */
+	public $ignoredAnnotationNames = [];
+
 	/**
 	 * @return mixed[]
 	 */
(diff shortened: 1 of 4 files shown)


SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
```diff
@@ -89,16 +92,18 @@ public function process(\PHP_CodeSniffer\Files\File $phpcsFile, $openTagPointer)
...
-
-							if ($matches[1] !== $nameAsReferencedInFile) {
-								/** @var \SlevomatCodingStandard\Helpers\Annotation $annotation */
-								foreach ($annotationsByName as $annotation) {
-									$phpcsFile->addError(sprintf(
-										'Case of reference name "%s" and use statement "%s" do not match.',
-										$matches[1],
-										$unusedNames[$uniqueId]->getNameAsReferencedInFile()
-									), $annotation->getPointer(), self::CODE_MISMATCHING_CASE);
+							if (!in_array($matches[1], $this->ignoredAnnotationNames, true)) {
```
kukulich: Is there a reason (I maybe missed something) why didn't you use the `$annotationName` variable? It can be used before the `preg_match` call.

tests/Sniffs/Namespaces/data/caseInsensitiveUse.php
```diff
@@ -11,7 +11,9 @@
 new BAR();
 
 /** @var boo */
-/** @BOO */
+/**
```
kukulich: Please add special annotation for your test case and don't change existing test cases.

SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
kukulich: Please use `SniffSettingsHelper::normalizeArray`.

xificurk:
@kukulich Thanks for the comments... it should be fixed now.

kukulich:
Thanks, squashed and cherrypicked: https://github.com/slevomat/coding-standard/commit/073671ebce6e9a33550247363ea6b2734c3ac23f

I made one small change. You have to set the ignored names with `@`. It's consistent with another our sniff.

xificurk:
Thank you :+1:

xificurk:
Well, that's kind of unfortunte. It means that I (and probably others) will need to either completely skip the 4.4.x branch or temporarily disable some sniffs and wait for the new release. But it's your call ;-) I'm happy that the fix is on its way.

kukulich:
New option is new feature so it has to wait for 4.5.0 release but I want to write some new sniffs to be in 4.5 as well.

dg:
Is possible to somehow ignore content of annotation? This code generaters error `Case of reference name "cache" and use statement "Cache" do not match`:

/--php
use Nette\Caching\Cache;

class FileStorage implements Nette\Caching\IStorage
{
	/** @internal cache file structure */
	const META_HEADER_LEN = 28;
\--

kukulich:
@dg It needs new option: https://github.com/slevomat/coding-standard/issues/293

PR #307, opened 2018-03-17
Fixed AppVeyor builds

kukulich:


Majkl578:
You may want to use Chocolatey instead to set up PHP on AppVeyor, see Doctrine DBAL config: https://github.com/doctrine/dbal/blob/v2.7.1/.appveyor.yml#L52

kukulich:
Yes, it’s on my todo list. Zip files are quicker but I really don’t know why are broken now :(

PR #309, opened 2018-03-20
Fix AbstractFullyQualifiedGlobalReference to implement sniff interface

MortalFlesh:
I guess that all Sniffs should implement sniff interface.


kukulich:
Cherrypicked, thanks.

PR #316, opened 2018-03-23
Update local Consistence CS copy

VasekPurchart:
Updated copy of Consistence CS after major rework of the XML.

I noticed two things we might adress in the PR too, though:

1) The `Generic.Arrays.ArrayIndent` is commented out in the copy, I am not sure why though, I have not encountered any problems probably apart from what was excluded?

2) I think it might better if the Slevomat sniffs were included here as well and only configured/added/removed where needed? I think it will be less confusing for updating the local ruleset. Or was there some technical block why we didn't do it like that before?

### DIFF
--- a/build/phpcs-consistence.xml
+++ b/build/phpcs-consistence.xml
@@ -14,13 +14,26 @@
 	<rule ref="Generic.CodeAnalysis.ForLoopShouldBeWhileLoop"/>
 	<rule ref="Generic.CodeAnalysis.UnconditionalIfStatement"/>
 	<rule ref="Generic.CodeAnalysis.UnnecessaryFinalModifier"/>
+	<rule ref="Generic.ControlStructures.InlineControlStructure"/>
+	<rule ref="Generic.Files.ByteOrderMark"/>
 	<rule ref="Generic.Files.InlineHTML"/>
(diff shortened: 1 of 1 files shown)


kukulich:
1. Not sure - there were probably some bug that has been solved in newer PHPCS version.
2. Not sure either :)

VasekPurchart:
Ok, I will try these both :)

VasekPurchart:
I have looked into those two situations:

1) I think it is ok, it discovered a code formatting problem :) #317
2) I think it was because of default values from Slevomat sniffs, if they are overridden by anything, you cannot go back to them and they would have to be configured explicitly in the local standard. I think that is fine, but you might have not liked it.  Tell me if you would prefer me to change it or not :)

kukulich:
@VasekPurchart 2. You can change it :)

VasekPurchart:
Prepared in #318, there is one more thing - whether you want to include all Slevomat sniffs by default - how it was done until now (I left it that way) or not.

Current situation means all sniffs - even those which are contradictory (like `DisallowYodaComparison` vs `RequireYodaComparison`), so everytime something must be excluded.

The other approach would be including only something (and only those, which are not part of Consistence CS).

kukulich:
Yes, it's better to include them.

PR #325, opened 2018-04-01
UnusedUseSniff: hotfix for ampersand in phpdoc

CZechBoY:
Hi,
this pr hotfixes intersection ampersand in phpdocs.
I can add test later.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
@@ -206,7 +206,7 @@ class UnusedUsesSniff implements \PHP_CodeSniffer\Sniffs\Sniff
 							}
 
 							foreach ($contentsToCheck as $contentToCheck) {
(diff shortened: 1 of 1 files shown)


kukulich:
I agree with @Majkl578

CZechBoY:
ok, but how do you solve it NOW?

Majkl578:
Use `|`, not `&`, it's not supported by IDEs and other tools anyway.

ikvasnica:
@Majkl578 When I use `|` instead of `&`, PHPStan will complain instead.

kukulich:
@ikvasnica It's a problem of PHPStan.

ikvasnica:
@kukulich There is a reason behing it, see: https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb
I think it makes perfect sense to use intersection types in case of creating mocks, for instance.

kukulich:
@ikvasnica Yes, but it is wrong when PHPStan enforces syntax that is not supported by IDEs and other tools. It should be enforced only if you enable special option or eg. via special `@phpstan-var` annotation.

PR #326, opened 2018-04-03
Make PSR 12 Compatible Sorting default in AlphabeticallySortedUsesSniff

simPod:
New IDEA 2018.1 seems to sort it in PSR 12 compatible way. I believe it's the same with PHPStorm.

Let's make it default


Majkl578:
:-1: 
This is huge BC break.

kukulich:
I may change it in Slevomat CS 5.0 but not now because it's a BC break.

simPod:
I actually like it so it's fine

Majkl578:
It actually breaks ton of code (even unrelated to this sniff), simply running _optimize imports_ breaks BC.

mmoll:
What about making it configurable? :)
https://github.com/mayflower/mo4-coding-standard/pull/44

PR #328, opened 2018-04-06
Disable caching before runing fixer

grongor:
The tests that are focused on fixer won't work if you have something like `<arg name="cache" value=".phpcs-cache"/>` in your ruleset.


kukulich:
Thanks.

PR #329, opened 2018-04-06
Make asserts static as rest of the methods

grongor:


kukulich:
Cherrypicked, thanks.

Issue #251, opened 2018-01-30
Parse annotations with phpstan/phpdoc-parser


kukulich:
https://github.com/phpstan/phpdoc-parser

Majkl578:
:top: Will this also add support for Foo<> notation?

kukulich:
Yes

CZechBoY:
Can we somehow ignore errors from `Foo<> notation `? But know whole sniff `SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.InvalidFormat`.

Majkl578:
Next PhpStorm EAP will support intersection types on parser level (but still treat them as unions), so this is more actual than ever. :)
https://youtrack.jetbrains.com/issue/WI-39419#focus=streamItem-27-3061342-0-0

Majkl578:
https://blog.jetbrains.com/phpstorm/2018/09/phpstorm-2018-3-eap-183-2635-12/

simPod:
PHPStan support is out https://github.com/phpstan/phpstan/commit/73cc2118c873c082aa8fbe7b037188ec4d04e7a1

kukulich:
Implemented in https://github.com/slevomat/coding-standard/commit/824a235ae9889eb274b353346ae653adace4b3f4 and https://github.com/slevomat/coding-standard/commit/91b8c2a816979b3480aeeda61a579450d7bfc974

PR #350, opened 2018-05-17
Fixes some more false-positives for UnusedUsesSniff

weirdan:
- Collection syntax (PSR-5): `Traversable<int,string>`
- Generics (psalm,phan): `Decorator<Classname>`
- Shapes (psalm): `@return array{length:int,data:string}`

Additionally, now corresponding regexp is built dynamically to make it
easier to extend.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
@@ -206,7 +206,12 @@ class UnusedUsesSniff implements \PHP_CodeSniffer\Sniffs\Sniff
 							}
 
 							foreach ($contentsToCheck as $contentToCheck) {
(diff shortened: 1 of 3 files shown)


kukulich:
I'm not sure if I want to merge this. I think it should be better to implement https://github.com/slevomat/coding-standard/issues/251

weirdan:
It would certainly be better, however I don't see it implemented anytime soon, and this sniff produces false-positives now. Some people (myself included) are holding up upgrades to newer versions because of this.

mavimo:
@kukulich I agree that the "best" solution is to migrate to new php parser, but I think it will take some time and we have the same issue reported from @weirdan, maybe this should be a good solution to allow users to migrate to the new version untile the 5.0.0 will be released?

Majkl578:
The problem with this _fix_ is that it introduces inconsistencies - now you are supporting _new_ syntaxes for unused uses, but not for other important sniffs. :|

kukulich:
It's not good solution for me as a developer of this library.

mavimo:
@kukulich I understand, there is something that I can do to help in the new parser implementation? Maybe there is already a discussion about it but I see only the issue, can you address me into it (if there is some mid-term plan)?

PR #361, opened 2018-06-11
Compatibility with PHPCS 3.3

kukulich:


Majkl578:
What a mess compared to 3.2.x. :cry:

kukulich:
It could be worse :) But it’s slower now

Issue #446, opened 2018-08-31
UseSpacing sniff breaks with multiple namespaces/uses in a file


Majkl578:
Doctrine usually has functional tests and its mocks/stubs for tickets in the same file. Ocassionally this could mean multiple namespaces and uses per file.

[This file](https://github.com/doctrine/doctrine2/blob/master/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2084Test.php#L66-L69) is such example, the sniff breaks (`FAILED TO FIX`) and reports:
```
FOUND 3 ERRORS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 68 | ERROR | [x] Expected 1 lines after last use statement, found 0. (SlevomatCodingStandard.Namespaces.UseSpacing.IncorrectLinesCountAfterLastUse)
 68 | ERROR | [x] Expected 0 lines between same types of use statement, found -2.
    |       |     (SlevomatCodingStandard.Namespaces.UseSpacing.IncorrectLinesCountBetweenSameTypeOfUse)
 69 | ERROR | [x] Expected 1 lines before first use statement, found 0. (SlevomatCodingStandard.Namespaces.UseSpacing.IncorrectLinesCountBeforeFirstUse)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 3 MARKED SNIFF VIOLATIONS AUTOMATICALLY
```
(notice `-2`)

Additionally the following code in the middle of file:
```php
use Doctrine\ORM\Annotation as ORM;
use Doctrine\Tests\OrmFunctionalTestCase;
```
is changed into:
```php
use Doctrine\ORM\Annotation as ORM;

use Doctrine\Tests\OrmFunctionalTestCase;


























```
And then still reported as failed to fix.

Majkl578:
Note that similar issue also affects other sniffs, i.e. `SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName` puts the `use` into wrong namespace.

kukulich:
I'll fix it in version 5.0.0. It's not possible without BC break.

Majkl578:
Ok, np, it only affects the fixer, check works fine IIRC after manually fixing the problem.

kukulich:
[...]

PR #455, opened 2018-09-04
Implement PHP.OptimizedFunctionsWithoutUnpacking sniff

### LINKED ISSUE #454: Sniff to check that engine-optimized functions do not use argument unpacking
There is [bunch of PHP functions](https://github.com/php/php-src/blob/850b53df0e3aa1a898db000244290fd0b7ecc8b6/Zend/zend_compile.c#L3916-L3975) that are optimized into specialized VM opcode. This [only works when argument unpacking is not used](https://github.com/php/php-src/blob/850b53df0e3aa1a898db000244290fd0b7ecc8b6/Zend/zend_compile.c#L3912-L3914).

Having a sniff which will point out when such function is used with variadics would be nice.

Majkl578:
closes #454

### DIFF
--- a/SlevomatCodingStandard/Sniffs/PHP/OptimizedFunctionsWithoutUnpackingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/PHP/OptimizedFunctionsWithoutUnpackingSniff.php
@@ -1,0 +1,137 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\PHP;
+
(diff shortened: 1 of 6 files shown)


kukulich:
It looks the version 4.8 would be necessary :)

Majkl578:
Split tests for namespaced and non-namespaced code into different files, seems to also be affected by #446.

Majkl578:
>  Coverage increased (+0.0001%) to 99.981%

:joy:

kukulich:
Perfect, thanks!

PR #471, opened 2018-09-11
Build: rename the phpunit config file

jrfnl:
PHPUnit offers the possibility to use a `.dist` file for project configurations.
Using that, will allow individual developers to overrule the project file with their own configuration which may, for instance, contain environment variables local to their system.


kukulich:
Will this work https://github.com/slevomat/coding-standard/blob/master/build.xml#L72 ?

jrfnl:
@kukulich That line should probably be removed as PHPCS will sort it out automatically if one of the recommended custom ruleset filenames is used. All the same, I haven't got a clue what that file is for, so I can't really guage whether it will work or not.

ondrejmirtes:
I think that `phpunit.xml.dist` and `phpunit.xml` should be in the root directory (or in the workdir) in order for PHPUnit to notice them.

jrfnl:
I've just adjusted the `build.xml` file for now, but you may want to look into letting PHPUnit (and PHPCS for that matter), sort these things out automatically.

I know for a fact that PHPCS will look in the directory it is run from and walk up the directory tree until it finds a custom ruleset file (unless a `--standard` is provided already).

kukulich:
Solved in https://github.com/slevomat/coding-standard/commit/b3d315df5dd7c0c40b173a78cc29bb2e14e1a0e3

PR #476, opened 2018-09-13
RequireShortTernaryOperatorSniff: infinite loop fix

dg:
Fix for this problem https://travis-ci.org/nette/di/jobs/427930497


kukulich:
Thanks!

PR #525, opened 2018-10-17
Fix README for TypeHintDeclarationSniff

gergelypap:
usefulAnnotations property is not deprecated.


kukulich:
`usefulAnnotations` is deprecated.

gergelypap:
Well, ok. Why though? I like the reporting on unnecessary @param, @return, etc annotations when there are PHP7 typehints available, but I don't want reporting on e.g. PHPUnit annotations such as @dataProvider.

Enabling `$allAnnotationsAreUseful` is not good for this type of checking. Am I missing something?

PR #539, opened 2018-10-28
Reproduce false positive for unused method use

ostrolucky:
This kind of code is used eg in Symfony ProgressBar


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/ae15ec0902f110c8ae36c07402383be12cfc49f3

PR #555, opened 2018-11-16
Fix ControlStructureSpacingSniff for cases when a control structure is at the very end of a file

echernyavskiy:
This one-liner is to fix a PHPCS ``Internal.Exception`` generated by ``ControlStructureSpacingSniff`` for a file like this:

    <?php
    
    try {
      true;
    }
    catch (\Exception $exception) {
    }

The PHPCS error is:

    An error occurred during processing; checking has been aborted. The error message was: Undefined index: in .../vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/ControlStructures/ControlStructureSpacingSniff.php on line 331 (Internal.Exception)

The issue is the ``try-catch`` construct being the last one in the file.


carusogabriel:
@echernyavskiy Maybe we should add a test to ensure this scenario?

kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/0c963c889c2cc69b821ccb9a25e2a5f9c1e28c55

PR #534, opened 2018-10-23
[WIP] Update PHPStan to 0.10

ondrejmirtes:

PR #558, opened 2018-11-22
[WIP] Start adding support for generics-style collections

thewilkybarkid:
(Very-much WIP!)

Starting to try and implement #251 and (BC) support for generics-style collections.

`TypeHintDeclarationSniffTest::testNoErrors` and `TypeHintDeclarationSniffTest::testErrors` (not yet added news cases) are green, other tests not yet fixed. The code changes are the result of hacking until it works(ish), so definitely a lot of tidying up to do as well. Opening this PR early just to get it visible.

Will probably result in a new-sniff to allow/force this syntax, but need to get it working properly first.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
 	},
 	"require": {
 		"php": "^7.1",
+		"phpstan/phpdoc-parser": "^0.2",
 		"squizlabs/php_codesniffer": "^3.3.1"
 	},
 	"require-dev": {
--- a/SlevomatCodingStandard/Sniffs/TestCase.php
+++ b/SlevomatCodingStandard/Sniffs/TestCase.php
@@ -7,6 +7,7 @@ use PHP_CodeSniffer\Config;
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Files\LocalFile;
 use PHP_CodeSniffer\Runner;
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -10,6 +10,7 @@
 	},
 	"require": {
 		"php": "^7.1",
+		"phpstan/phpdoc-parser": "^0.2",
```
thewilkybarkid: Can't use `^0.3` until https://github.com/slevomat/coding-standard/pull/534 is complete.

thewilkybarkid: Haven't checked what changed... hopefully isn't much...

kukulich:
@thewilkybarkid Thank you for yourt effort but I think the implementation of the parser should be done by me because I need to figure out how to use it in all sniffs.

thewilkybarkid:
@kukulich Do you know when you would get around to it? This is looking like it could be done in a BC way; I'm happy to commit time over the next week to try and finish it (and there could be a major release after to tidy it up), but won't if you wouldn't merge it.

kukulich:
I want to fix currently known bugs, make a bugfix release a then work on this. I don't have any timeline because I don't have any time currently.

This PR will not be merged because I want the support in all sniffs at once.

PR #560, opened 2018-12-02
Fixed recognizing reference names as these could contain spaces

webimpress:
We can have namespace with spaces in PHP. This is completely valid code with spaces around namespace separator:

```php
namespace Hello \ World;

use MyName\ Space;

class Hey extends \ Other \ ParentClass
{
    use MyTraits
        \Aware;
}
```

etc.

Recognising references with spaces doesn't work correctly now which causes some issues where using with other sniffs. We are getting imported wrong class in first iteration I've seen something like `use ;` (without class name) - what is invalid code and breaks other sniffs in the iteration.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/876fde0ad8f07700c911e0f628ca80eb460f3dff

PR #582, opened 2019-01-08
ReferenceUsedNamesOnlySniff: Check @method annotations

hranicka:
Hello,

rule `SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly` does not check `@method` annotations when `searchAnnotations` option is set. So codebase still contains FQN in this annotation despite the coding standard forbids that.

This PR adds `@method` annotation to the list of checked ones.


kukulich:
Thanks.

PR #595, opened 2019-01-22
Add test case for invalid useless variable report

grongor:
`UselessVariableSniff` incorrectly reports that a variable is useless there. If you change the name of either the static variable or the local one, the sniffs reports no errors.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/65d8048cafee8d6008c93bf8c72c7ce6c16890c8

PR #601, opened 2019-02-06
Follow type hint when converting to generics syntax

thewilkybarkid:
Thanks for the work implementing generics. 🎉

Found a bug try it out in https://github.com/libero/php-coding-standard/pull/40: argument/return type hints aren't used, so it always becomes `array` rather than `iterable`, `ArrayObject` etc.

### DIFF
--- a/tests/Sniffs/TypeHints/data/disallowArrayTypeHintSyntaxErrors.fixed.php
+++ b/tests/Sniffs/TypeHints/data/disallowArrayTypeHintSyntaxErrors.fixed.php
@@ -42,4 +42,20 @@ class Whatever
 	/** @var array<int>|string */
 	public $k;
 
+	/** @return iterable<array<string>> */
+	public function l() : iterable {
+	}
+
+	/** @param iterable<array<string>> $n */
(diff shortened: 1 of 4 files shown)


tests/Sniffs/TypeHints/data/disallowArrayTypeHintSyntaxErrors.fixed.php
```diff
@@ -42,4 +42,20 @@ class Whatever
 	/** @var array<int>|string */
 	public $k;
 
+	/** @return iterable<array<string>> */
```
thewilkybarkid: Followed https://github.com/slevomat/coding-standard/blob/fdab228c1800db9e65a62dde306bfa856d98fcc3/tests/Sniffs/TypeHints/data/disallowArrayTypeHintSyntaxErrors.fixed.php#L30 in keeping nested items as `array`.

kukulich:
@thewilkybarkid Thank you. I've implemented the sniff very easy by purpose and wait for response :) So thank you for the response :)

PR #626, opened 2019-03-06
Cover short ternaries in UselessTernaryOperatorSniff

thewilkybarkid:
Rather an edge case!


thewilkybarkid:
AppVeyor failure is unrelated.

Majkl578:
Can you provide some real-world use case for this? I'd be really interested in seeing one.

thewilkybarkid:
Heh, so would I! Just spotted it when adding a (rather unrealistic) test case so our CS (https://github.com/libero/php-coding-standard/pull/50/files#diff-abd230f3b7a1a6e0e56045b82f054108R9). Reasonably quick and should be harmless to add though.

kukulich:
@thewilkybarkid Please squash commits to one commit and name it in the same way as commits in master.Thank you.

kukulich:
Thanks.

PR #627, opened 2019-03-06
Don't throw exceptions when something isn't supported

thewilkybarkid:
Currently exceptions are thrown, which isn't particularly useful as it breaks the process. Instead, this just skips them, so that it works with `Generic.ControlStructures.InlineControlStructure`. (Haven't looked into whether it's possible to add support instead, but this should be enough.)


thewilkybarkid:
AppVeyor failure is unrelated.

kukulich:
I'm sorry but I don't like to silently ignore it so I will not merge it.

thewilkybarkid:
@kukulich Would you be willing to have an option to skip instead?

PR #651, opened 2019-03-31
Fix return type hint finder [BUG]

Billz95:
The `FuncitonHelper`'s `findReturnTypeHint` is not working properly under the following condition:

```php
function foo(): array{
}
```
but works with the following code
```php
function foo(): array {
}
```

The only difference is a space after the `array` type hint. 

This was due to the off-by-one parameter of the `TokenHelper::findeNextExcluding` function. 

The last parameter `endPointer` should be exclusive, however, when calling this function, a `-1` was used explicitly to specify the exclusiveness.

After removing the `-1`, everything seems to be working fine.


Billz95:
AppVeyor build failed due to some php7.3 installation issue, doesn't seem to be caused by this PR..

kukulich:
Thank you for your report.

Cherry-picked https://github.com/slevomat/coding-standard/commit/649fdaa10caea08479f0e6039ae0543f879cf9ff

PR #657, opened 2019-04-04
UnusedUsesSniff: Better handle php open tags

donatj:
UnusedUsesSniff gets false positives if more than one `<?php` opening tag is used in a file. This is because the sniff was defining the scope of a namespace to the _most recent_ T_NAMESPACE or *most recent* T_OPEN_TAG.

This was pretty easily resolved by changing the definition of the namespace to the _most recent_ T_NAMESPACE or the **first** T_OPEN_TAG.


kukulich:
Thanks.

Issue #633, opened 2019-03-14
Split ControlStructureSpacingSniff or make it more configurable


grongor:
I would like to configure phpcs to check blank lines only between and after the control structures for cases like `if`, `foreach`, `while`, ... and to require blank line before jump statements `continue`, `return`, ... so that the code like this would be valid:

```php
$things = $this->getBlabla();
foreach ($things as $thing) {
    if ($thing === 1) {
        continue;
    }

    if ($thing === 2) {
        break;
    }

    if ($thing === 3) {
        $x = 1;

        continue;
    }

    break;
}
```

And this code would "autofix" to the previous one:
```php
$things = $this->getBlabla(); // there might or might not be a blank line before, we don't care
foreach ($things as $thing) {
    if ($thing === 1) {

        continue;
    }
    if ($thing === 2) {
        break;
    }
    if ($thing === 3) {
        $x = 1;
        continue;
    }
    break;
}
```

Would you consider accepting such changes if I implement them? Maybe even split the sniff into two (`if`s vs `continue`s)?

kukulich:
@grongor How would you name the new sniffs? And what tokens will they support?

grongor:
I think the two groups should be:
```
T_IF,
T_DO,
T_WHILE,
T_FOR,
T_FOREACH,
T_SWITCH,
T_TRY,
```

```
T_GOTO,
T_BREAK,
T_CONTINUE,
T_RETURN,
T_THROW,
T_YIELD,
T_YIELD_FROM,
```

And named maybe `ControlStructures` and `JumpStatements`?

kukulich:
@grongor The first should be `BlockControlStructures` or something similar so I can release version 5.1.0 with the two new sniffs and with `ControlStructureSpacingSniff` marked as deprecated.

`BlockControlStructures` can probably only extend `ControlStructureSpacing` with different defaults and throwing exception for unsupported tokens.

So yes, I will definitely merge it :)

grongor:
Ok thanks, I'll send PR soon ;-) thx

kukulich:
@grongor Will you have time to send the PR? :)

grongor:
@kukulich sorry, I had other priorities this month. I will look into it next week :)

kukulich:
@grongor Understood :) Thanks.

grongor:
@kukulich ok, done. So sorry for the delay - getting back from vacation was harder then I expected :joy: 

There are two new things:
[...]

PR #686, opened 2019-05-15
Split ControlStructureSpacingSniff into two separe sniffs

grongor:
Also increase the configurability of both sniffs

fixes https://github.com/slevomat/coding-standard/issues/633


kukulich:
@grongor Thank you!

First quick review:
- There are broken CS, indentation especially
- Please don't extend `BlockControlStructureSpacingSniff` in `JumpStatementsSpacingSniff`. If there is much common, make one abstract sniff, something like `AbstractFullyQualifiedGlobalReference` 
- Make `ControlStructureSpacingSniff` parent of `BlockControlStructureSpacingSniff` and mark it as deprecated. You have to solve also compatibility of `linesCountAroundControlStructure` with the new two options.
- Please update README
- Please remove `-1` as possible value. It's better to just ignore right sniff codes.

Thank you again :)

grongor:
- Fixed the CS etc already, forgot to run it locally :)
- I thought you wanted it this way ... but yeah, sure, I can extract the stuff to an abstract parent
- sure, makes sense
- sure
- oh yeah, of course, I didn't realize that I can achieve the same thing with ignoring the sniff code...

I'll all fix it tomorrow.

kukulich:
> I thought you wanted it this way ... but yeah, sure, I can extract the stuff to an abstract parent

https://github.com/slevomat/coding-standard/issues/633#issuecomment-474012790

:)

grongor:
@kukulich done :)

grongor:
@kukulich I've added one more option to allow yield stacking, I hope you like it. I've stumbled upon "an issue" with code likes this:
```php
private function getSensorStates(Device $device, array $oidSensorMap) : iterable
{
    yield from $this->getStatesValues($device, $oidSensorMap);
    yield from $this->getProcessorLoadValues($device, $oidSensorMap);
    yield from $this->getDeviceStatusValues($device, $oidSensorMap);
    yield from $this->getStorageSizeValues($device, $oidSensorMap);
    yield from $this->getStorageUsedValues($device, $oidSensorMap);
    yield from $this->getPhysicalEntityValues($device, $oidSensorMap);
}
```

and I think that adding the blank lines here serves no purpose ... hence the new option.

kukulich:
@grongor Yes, no problem with this new option. Should I review now? :)

grongor:
> I think this option should be only in JumpStatementsSpacingSniff

Yeah, I wasn't quite sure where to put it ... you mean only the option, or somehow move the check to this sniff too?

grongor:
@kukulich yeah sure, I'm done I guess :D Just be more specific about what exactly to do with the new option/code, otherwise I'll wait for your review ;-)

README.md
```diff
@@ -327,6 +328,12 @@ For example, with the following setting, only `if` and `switch` tokens are check
 </rule>
 ```
 
+#### SlevomatCodingStandard.ControlStructures.JumpStatementsSpacing 🔧
+
+Enforces configurable number of lines around jump statements (continue, return, ...).
+
+See `SlevomatCodingStandard.ControlStructures.BlockControlStructureSpacing` for more information (same parameters apply).
```
kukulich: Please copy the options sections. And the new option `allowSingleLineYieldStacking` :)

SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacingSniff.php
```diff
@@ -0,0 +1,388 @@
...
+	 * @return int[]
+	 */
+	abstract protected function getDefaultTokensToCheck(): array;
+
+	/**
+	 * @return (int|string)[]
+	 */
+	private function getTokensToCheck(): array
+	{
+		if ($this->normalizedTokensToCheck === null) {
```
kukulich: We should use `getDefaultTokensToCheck()` to also filter only supported tokens. So maybe the name `getSupportedTokens` would be better.

SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacingSniff.php
```diff
@@ -0,0 +1,388 @@
...
+	public $linesCountBeforeFirstControlStructure = 0;
+
+	/** @var int */
+	public $linesCountAfterControlStructure = 1;
+
+	/** @var int */
+	public $linesCountAfterLastControlStructure = 0;
+
+	/** @var bool */
+	public $allowSingleLineYieldStacking = true;
```
kukulich: Move this option to `JumpStatementsSpacing` please. I'm ok if you move only the option :)

SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacingSniff.php
grongor: You mean to now allow other tokens then the default ones? That would break one test with `switch`, where custom tokens `T_CASE` and `T_DEFAULT` are used ...

Wouldn't it be better to leave it "open"? If it works for users to use different kinds of tokens, so be it...

kukulich: I don't want reports that something with custom tokens is not work as expected. 

So we need `getSupportedTokens` and also `getDefaultTokensToCheck()` because of `T_CASE` and `T_DEFAULT`.

SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacingSniff.php
grongor: Now that I think about it it would be better to move it all ... I'm also not sure how would I separate the option and the code :D

SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacingSniff.php
grongor: Oh ok, so `T_CASE` and `T_DEFAULT` should be supported but not checked by default ... got it

kukulich: It looks there another option :)

The support for `T_CASE` and `T_DEFAULT` was added later and are not in default tokens because of backward compatibility.

So we can have only `getSupportedTokens`. `T_CASE` and `T_DEFAULT` could be among supported tokens in `BlockControlStructureSpacingSniff` and not among in the old `ControlStructureSpacingSniff`

grongor: Ah, I think I'm getting a bit lost here :D Check the latest commit and we will see... :D

kukulich: @grongor It's ok, I will merge it :)

Please squash and rebase.

grongor:
Hi, any ETA on tagging a release with this? :)

PR #687, opened 2019-05-16
Fix implementation of method findFirstTokenOnNextLine

grongor:
Before changing the implementation I was getting this test result:

```
1) SlevomatCodingStandard\Helpers\TokenHelperTest::testFindFirstTokenOnNextLineEndingWithAComment
Expected T_STRING, actual token is T_VARIABLE
Failed asserting that 320 is identical to 319.
```

The problem is that new line is part of the `T_COMMENT` and the content comparison in the `findNext` method is done "exactly" (`$content === $tokens[$i]`), therefore the new line in the comment is not detected (the token is skipped).


kukulich:
Thank you.

PR #688, opened 2019-05-17
Fixes issue when getCommentContent returns false

nunomaduro:
Makes sure a comment always haves a string content.


kukulich:
@nunomaduro Could je please post test case ?

nunomaduro:
Sure, origin: https://github.com/nunomaduro/phpinsights/issues/62.

Unterminated comments:
```php
/*

Route::get('/', function () {
    return view('welcome');
});
```

kukulich:
I don't know if we need to support this...

nunomaduro:
You don't need to support this, maybe we could adapt to this situation as this is not an empty comment.

ttrig:
This will also take care of files ending with e.g. `/* Line comment */` where the last character is not a newline. :+1:

nunomaduro:
And tests are passing, so. 🤗

kukulich:
For me it's invalid code. I don't I should solve this.

francislavoie:
@kukulich if PHP runs it, it's valid code. This tool can end up being used on legacy apps with totally varying degrees of code quality. When trying to use it with a tool like https://github.com/nunomaduro/phpinsights to assess the code quality, this issue makes things break.

kukulich:
It may be valid code but this repository does need to support all valid codes (and already does not). The code is invalid for me.

nunomaduro:
Alright, what can we make to do things better here? Throwing a fatal exception don't seems right to me.

francislavoie:
To clarify: `getCommentContent` declares a return type of `string`. `substr` returns a string, or false on failure. Simply returning the output of `substr` breaks the return type contract when `substr` returns false. This PR corrects that behaviour and prevents an erroneous fatal from occurring due to the type checker. I don't see what's controversial about that.

genesiscz:
> It may be valid code but this repository does need to support all valid codes (and already does not). The code is invalid for me.

How should I even know what's wrong with my old project and where is this document wrong if only think I get is just a fatal error without any explanation.  Since it's kinda valid code for PHP, I guess other checkers won't find this as an issue. How am I supposed to track this?

genesiscz:
Also, this is PERFECTLY valid code:

/**
 something here /**
*/

PHP won't get mad, this package does. Found out by using -v (verbose) for phpcs

PR #689, opened 2019-05-20
TypeNameMatchesFileNameSniff: failing test for file without namespace

xificurk:
`TypeNameMatchesFileNameSniff` completely ignores files without namespace declaration.

I think the expected behaviour is to report an error.


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/e7c3f86225284f37b41bc4a5cef2161d8bc3fc19

I really don't know why we didn't report it by default...

xificurk:
Thank you for the quick fix :+1:

PR #694, opened 2019-05-22
Composer: update the PHPUnit requirement

### LINKED ISSUE #693: Composer installation conflict
There is an inconsistency/discrepancy in the `composer.json` file.

In the `require` section it lists `"php": "^7.1",`, but when running `composer install --no-dev` on PHP 7.1.x, it will not install as Composer will still look at `require-dev` and PHPUnit 8 has a minimum PHP requirement of PHP 7.2.

```
Your requirements could not be resolved to an installable set of packages.

Problem 1
  - Installation request for phpunit/phpunit 8.1.4 -> satisfiable by phpunit/phpunit[8.1.4].
  - phpunit/phpunit 8.1.4 requires php ^7.2 -> your PHP version (7.1.25) does not satisfy that requirement.
  
Running update with --no-dev does not mean require-dev is ignored, it just means the packages will not be installed. If dev requirements are blocking the update you have to resolve those problems.
```

jrfnl:
As the unit tests work perfectly fine on PHPUnit 7.5+, allow for PHPUnit 7 to be installed to prevent the Composer version conflict.

Tested & found working.

Fixes #693

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
 		"phpstan/phpstan": "0.11.6",
 		"phpstan/phpstan-phpunit": "0.11.1",
 		"phpstan/phpstan-strict-rules": "0.11",
-		"phpunit/phpunit": "8.1.4"
+		"phpunit/phpunit": "^7.5 || ^8.1"
 	},
 	"autoload": {
 		"psr-4": {


kukulich:
Thank you but I will not downgrade PHPUnit.

jrfnl:
@kukulich I think you misunderstand. The current tests work fine with PHPUnit 7.5 _as well as_ PHPUnit 8.x, so nothing needs to change or be downgraded in the tests.

This PR solves an actual problem - i.e. that while this repo states that the minimum PHP requirement is PHP 7.1, **_you cannot install it on PHP 7.1 as `composer install` will fail_**, so I don't understand why you are closing the PR without even giving it proper consideration.

kukulich:
@jrfnl I'm sorry for the short answer.

See https://travis-ci.org/slevomat/coding-standard/jobs/535611213
It installed PHPUnit 7.5.0. I really don't want to solve some future inconsistencies in my builds because of old PHPUnit.

kukulich:
Btw you can use `--ignore-platform-reqs`

jrfnl:
> It installed PHPUnit 7.5.0. I really don't want to solve some future inconsistencies in my builds because of old PHPUnit.

I can understand that, but while those potential future inconsistencies don't exist yet and may never come to pass, there is no reason to change the minimum PHP requirements of the repo.

If/when in the future, those inconsistencies would come to pass, you will see it immediately as that Travis job will fail and _that_ is the moment to drop PHP 7.1 and PHPUnit 7.x support, not now.

kukulich:
PHP 7.1 is already dead https://www.php.net/supported-versions.php

I was thinking earlier to remove the support (I don't test it already) so now I have a reason :)

jrfnl:
> PHP 7.1 is already dead https://www.php.net/supported-versions.php

Oh, I know, that doesn't mean that everyone has embraced the new & shiny yet.

> I was thinking earlier to remove the support (I don't test it already) so now I have a reason :)

Fair enough.

Just so you know, this just makes it a hell of lot more difficult to include this package in projects which do still support older PHP versions.

I'm trying to figure out a solution for one such project now, minimum PHP requirement of the project is PHP 5.6, they run the sniffs on 7.3, so there's no problem there, however as soon as this package is added to the `require-dev`, they will have to use `--ignore-platform-reqs` for non-dev which is pretty annoying.

PR #696, opened 2019-05-22
HOTFIX: Fix compatibility with non-Composer installs of PHPCS

### LINKED ISSUE #632: Fatal error: TokenIterator not found
We are facing an issue after updating to 5.x here:

> PHP Fatal error:  Uncaught Error: Class 'PHPStan\PhpDocParser\Parser\TokenIterator' not found in /home/travis/build/spryker/spryker/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationHelper.php:249

Locally this runs fine, but on travis this happens.

We have the same composer install there, working fine. And the same versions as expected:

  - Installing phpdocumentor/graphviz (1.0.4): Loading from cache
  - Installing squizlabs/php_codesniffer (3.4.0): Loading from cache
  - Installing phpstan/phpdoc-parser (0.3.1): Downloading (100%)
  - Installing slevomat/coding-standard (5.0.2): Downloading (100%)
  - Installing spryker/code-sniffer (dev-master eacf253): Downloading (100%)

The command is as always:

> vendor/bin/phpcs Bundles --standard=vendor/spryker/code-sniffer/Spryker/ruleset.xml -p

### LINKED ISSUE #695: Fatal error: Class 'PHPStan\PhpDocParser\Parser\TokenIterator' not found
Following up on #632 which reported the same error and was unjustly dismissed. /cc @dereuromark

This *is* an issue with the Slevomat standard and should be solved here.

Setup:
* PHP 7.2.13
* SlevomatCS git clone with branch `master` checked out.
    - Ran `composer install --no-dev`.
    - As Phing is not available on no-dev, did not run `bin/phing`
* PHPCS git clone with branch `master` checked out.
    - `phpcs` command available globally.
    - Added the path to SlevomatCS to the PHPCS `installed_paths`.

When I run the following command on a random code directory (to test which sniffs I may want to add to the ruleset for the project):
`phpcs . --extensions=php --basepath=. --tab-width=4 --standard=SlevomatCodingStandard`

... I receive the following fatal error:
```bash
# phpcs . --extensions=php --basepath=. --tab-width=4 --standard=SlevomatCodingStandard
PHP 7.2.13 (cli) (built: Dec  5 2018 07:34:22) ( ZTS MSVC15 (Visual C++ 2017) x86 )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies

PHP_CodeSniffer version 3.5.0 (stable) by Squiz (http://www.squiz.net)

Fatal error: Uncaught Error: Class 'PHPStan\PhpDocParser\Parser\TokenIterator' not found in /path/to/slevomat/SlevomatCodingStandard/Helpers/AnnotationHelper.php:245
Stack trace:
#0 /path/to/slevomat/SlevomatCodingStandard/Helpers/AnnotationHelper.php(224): SlevomatCodingStandard\Helpers\AnnotationHelper::parseAnnotationContent('@param', 'string $ext Ext...')
#1 /path/to/slevomat/SlevomatCodingStandard/Sniffs/Commenting/DocCommentSpacingSniff.php(124): SlevomatCodingStandard\Helpers\AnnotationHelper::getAnnotations(Object(PHP_CodeSniffer\Files\LocalFile), 151)
#2 /path/to/PHP_CodeSniffer/src/Files/File.php(489): SlevomatCodingStandard\Sniffs\Commenting\DocCommentSpacingSniff->process(Object(PHP_CodeSniffer\Files\LocalFile), 151)
#3 /path/to/PHP_CodeSniffer/src/Files/LocalFile.php(91): PHP_CodeSniffer\Files\File->process()
#4 /path/to/PHP_CodeSniffer/src/Runner.php(630): PHP_CodeSniffer\Files\LocalFile->process()
#5 /path/to/PHP_CodeSniffer/src/Runner.php(434): PHP_CodeSniffer\Runner->processFile(Object(PHP_CodeSniffer\Files\LocalFile))
#6 /path/to/PHP_CodeSniffer/src/Runner.php(114): PHP_CodeSniffer\Runner->run()
#7 /path/to/PHP_CodeSniffer/bin/phpcs(18): PHP_CodeSniffer\Runner->runPHPCS()
#8 {main}
  thrown in /path/to/slevomat/SlevomatCodingStandard/Helpers/AnnotationHelper.php on line 245

Exit code is 255
```

The issue is really easy to reproduce:
1. Install the latest PHPCS Phar file: https://github.com/squizlabs/PHP_CodeSniffer/releases/download/3.4.2/phpcs.phar
2. Clone this repo and run `composer install --no-dev`.
3. Register the standard with PHPCS `path/to/phpcs.phar --config-set installed_paths /path/to/slevomat-clone`
4. Now from the root of this repo run: `path/to/phpcs.phar . --extensions=php --tab-width=4 --standard=SlevomatCodingStandard`

I've ran some tests:
* This happens on all release tags in the 5.x series.
* `git bisect` shows me that commit 824a235ae9889eb274b353346ae653adace4b3f4 is to blame.
* Running `composer install` (=dev) & `bin/phing` makes no difference. Problem still occurs.

Now, you can blame PHPCS's autoloader, but that is neither here nor there. The simple fact is that this is broken.

If PHPCS detects itself to be installed in the `vendor` directory, PHPCS's autoloader unloads the Composer autoload file to place itself _before_ it and then will re-register the Composer autoload file (if found) after its own autoloader, so the Composer autoload is still available.

PHPCS also offers the ability to add a custom autoloader if you'd need it.

Now, I've done some further analysis.

PHPCS offers four different ways to install the tool:
1. As a Phar.
2. Via PEAR.
3. As a git clone.
4. As a Composer dependency of project.

External standards can normally be installed via [3] or [4] and need to be registered with PHPCS via `installed_paths`.

What it comes down to is that the Slevomat standard will currently **_only_** work when PHPCS is installed via [4] and run via `vendor/bin/phpcs` in a project or `bin/phpcs` when in the Slevomat standard itself. 

**Running the Slevomat standard with any of the other (supported) ways to install PHPCS is currently broken** as the Slevomat standard expects the Composer `autoload.php` file to be included while for any of the other methods it won't automatically be.

This is not a shortcoming of PHPCS itself. This is an issue with the Slevomat standard which clearly never took any of the other install methods into account.

The solution is to register your own autoloader which will load the Composer autoload file if necessary.

jrfnl:
For a full explanation of what was going on and why this will fix it, please read issue #695.

Fixes #695 
Fixes #632

### DIFF
--- a/SlevomatCodingStandard/ruleset.xml
+++ b/SlevomatCodingStandard/ruleset.xml
@@ -1,2 +1,4 @@
 <?xml version="1.0"?>
-<ruleset name="Slevomat Coding Standard"/>
+<ruleset name="Slevomat Coding Standard">
+	<autoload>./../autoload-bootstrap.php</autoload>
+</ruleset>
--- a/autoload-bootstrap.php
+++ b/autoload-bootstrap.php
@@ -1,0 +1,11 @@
+<?php declare(strict_types = 1);
+
+if (defined('SLEVOMAT_PHPCS_AUTOLOAD_SET') === false) {
+
+	// Check if this is a stand-alone installation.
+	if (is_file(__DIR__ . '/vendor/autoload.php')) {
+		require_once __DIR__ . '/vendor/autoload.php';
+	}
+
+	define('SLEVOMAT_PHPCS_AUTOLOAD_SET', true);
+}
--- a/build/phpcs.xml
+++ b/build/phpcs.xml
@@ -18,6 +18,10 @@
 		<exclude name="SlevomatCodingStandard.Namespaces.FullyQualifiedClassNameAfterKeyword"/>
 	</rule>
 
+	<rule ref="PSR1.Files.SideEffects.FoundWithSymbols">
+		<exclude-pattern>autoload-bootstrap.php</exclude-pattern>
+	</rule>
+
 	<rule ref="Generic.Commenting.Todo"/>
 	<rule ref="Squiz.PHP.CommentedOutCode">
 		<properties>


jrfnl:
> Please add this file to build.xml (tasks cs and cs-fix so CS are checked and fixed

Sorry, but why are you passing those command-line arguments via `build.xml` (in duplicate) instead of setting them in the ruleset ?
I presume you have a good reason, but this really doesn't make sense to me.

> Please fix CS

Will do. I was already surprised I wasn't getting any issues :smile: 

> Please don't use file_exists, use is_file

 Will do.

> Please rebase on master.

Will do.

kukulich:
@jrfnl 

> Sorry, but why are you passing those command-line arguments via build.xml (in duplicate) instead of setting them in the ruleset ?
I presume you have a good reason, but this really doesn't make sense to me.

Feel free to change it in another commit :) It's just historical and no need to change it :)

jrfnl:
I've opened #698 with the configuration changes. Once that's merged, I will rebase this PR and if you're ok with it, force push the updated commit.

jrfnl:
Rebased, updated & force pushed.

I've had to add an exclusion for the `PSR1.Files.SideEffects` sniff though as otherwise we can't use the constants to prevent the autoloader from potentially being registered multiple times.

dereuromark:
Thank you for fixing this :)

jrfnl:
Oh, before I forget: this fix will only work when the `SlevomatCodingStandard` ruleset is read out by PHPCS.

What that means is:
* This will work fine: `phpcs . --standard=SlevomatCodingStandard`
* This will also work fine: `phpcs . --standard=SlevomatCodingStandard --sniffs=slevomatcodingstandard.namespaces.unuseduses,slevomatodingstandrd.namespaces.uselessalias,slevomatcodingstandard.namespaces.fullyqualifiedconstants,slevomatcodingstandard.namespaces.fullyqualifiedglobalfunctions`
* Just like this:
    ```xml
    <?xml version="1.0"?>
    <ruleset name="Project">
        <rule ref="SlevomatCodingStandard"/>
    </ruleset>
    ```
* And this:
    ```xml
    <?xml version="1.0"?>
    <ruleset name="Project">
        <rule ref="SlevomatCodingStandard">
            <exclude name="SlevomatCodingStandard.Namespaces.RequireOneNamespaceInFile"/>
            <exclude name="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly"/>
            ...
        </rule>
    </ruleset>
    ```

But this will not work out of the box:
```xml
<?xml version="1.0"?>
<ruleset name="Project">
    <rule ref="SlevomatCodingStandard.Namespaces"/>
</ruleset>
```

Nor will this:
```xml
<?xml version="1.0"?>
<ruleset name="Project">
    <rule ref="SlevomatCodingStandard.PHP.TypeCast"/>
    <rule ref="SlevomatCodingStandard.Namespaces.FullyQualifiedGlobalConstants"/>
</ruleset>
```

In both those cases, the custom ruleset would need to add the Slevomat autoload to their own ruleset, like so:
```xml
<?xml version="1.0"?>
<ruleset name="Project">
    <!-- Note: path can be relative or absolute, both will work. -->
    <autoload>/path/to/slevomat/autoload-bootstrap.php</autoload>
    <rule ref="SlevomatCodingStandard.Namespaces"/>
</ruleset>
```

This is a known issue - see: https://github.com/squizlabs/PHP_CodeSniffer/issues/1582

PR #697, opened 2019-05-22
PHPUnit: various tweaks

jrfnl:
* `.gitattributes` - same as the `tests` directory, the `phpunit.xml.dist` config file does also not need to be included in the packaged release.
* `.gitignore` - ignore the standard PHPUnit overload config file which may be used by individual devs.
* `phpunit.xml.dist` - validated the XML file and added the schema header.
* `phpunit.xml.dist` - add the testsuite node so the `phpunit` command can be run without additional command-line parameters.
* `build.xml` - remove the now unnecessary `path="tests"` command line argument.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,7 @@
 .gitattributes export-ignore
 .gitignore export-ignore
 /.travis.yml export-ignore
+/phpunit.xml.dist export-ignore
 /build export-ignore
 /build.xml export-ignore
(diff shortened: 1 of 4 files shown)


kukulich:
Thank you, cherry-picked with some changes https://github.com/slevomat/coding-standard/commit/3cbb0240dfb5c014bebb30d8ada24912f81851ca

PR #700, opened 2019-05-22
PHPCS: add schema tags to the ruleset

jrfnl:
See:
* https://github.com/squizlabs/PHP_CodeSniffer/pull/1433
* https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.2.0

Note: I have not done this for the `Consistence` ruleset as that is a copy of another ruleset and should be fixed there.


kukulich:
Thanks.

PR #702, opened 2019-05-23
Fix JumpStatementsSpacingSniff

grongor:
Next semicolon was guessed incorrectly when the argument to return/throw/yield was a function call which itself had an array and a callback arguments.


grongor:
... ok, it broke something else ... gonna fix it, gimme a second

grongor:
Nah, not gonna finish it today, I'll post a fix (without another bug, hopefully) tomorrow

Issue #739, opened 2019-09-06
SlevomatCodingStandard.Functions.UselessParameterDefaultValue does not understand variadic parameters


dakorpar:
for example 

```php
...
public function __construct(string $a = 'something', string $b = 'b', ...$parameters); 
//this triggers error and it will be changed to
```
```php
public function __construct(string $a, string $b, ...$parameters); 
// this is never wanted.
```

kukulich:
https://github.com/slevomat/coding-standard/pull/740#issuecomment-547326884

PR #740, opened 2019-09-06
UselessParameterDefaultValueSniff should include variadic parameters in check

dakorpar:
fixes:
https://github.com/slevomat/coding-standard/issues/739

not sure what exactly variable_length means and how to cover that, but this wasn't covered before, this is better covered then before this PR.


kukulich:
CNR: https://github.com/slevomat/coding-standard/commit/31a54af1a5d6896e0c45d4e59af474f4da70f774

PR #748, opened 2019-09-26
Use reset to move to start of array, not current()

TysonAndre:
Fix use of current.://www.php.net/current

> The current() function simply returns the value of the array element
> that's currently being pointed to by the internal pointer.
> **It does not move the pointer in any way.**

Detected with PhanPluginUseReturnValueInternalKnown (checks for unused results of functions without side effects)

This probably wouldn't matter in php 7, unless something else was also calling functions such as next() and end()


kukulich:
Thank you.

PR #749, opened 2019-09-26
Remove unused arrays.

TysonAndre:
The array wasn't used anywhere else.
A different variable created with array_flip seems to have replaced the
original check.

Detected with PhanUnusedVariable. Tests passed locally.


kukulich:
Thank you.

PR #779, opened 2019-12-01
php 7.1 and 7.2 backward compatibility.

SimoTod:
This PR removes a trailing comma in a function call (this feature was introduced in php 7.3) to ensure backward compatibility with php 7.1 and 7.2 as stated in the composer.json file.
It should fix the errors reported by travis.


kukulich:
Thanks.

PR #784, opened 2019-12-04
Composer update

SimoTod:
The new version of phpcs supporting arrow function has been released. It would be good to use the tagged version in composer json so it will be less likely to conflict with other project stability requirements.


kukulich:
Both is already in master.

PR #817, opened 2020-01-08
Fix yield inside function calls

grongor:
Hi, here is a fix for https://github.com/cdn77/coding-standard/issues/21


kukulich:
Thanks.

PR #821, opened 2020-01-11
Fix crash in RequireCombinedAssignmentOperatorSniff

TysonAndre:
This causes an uncaught TypeError for malformed php files such as:

```php
<?php
$x =
```


kukulich:
I'm sorry but PHPCS really needs valid PHP.

Issue #826, opened 2020-01-14
RequireTernaryOperatorSniff false positive on reference


sebastienbarre:
The following code triggers " Use ternary operator. "

```php
    $condition = true;
    $foo = 1;
    $bar = 2;

    if ($condition) {
      $ref = &$foo;
    } else {
      $ref = &$bar;
    }
```

However, it is illegal to use a ternary operator with refs in PHP:
```
    $ref = $condition ? &$foo : &$bar;
```
will trigger "unexpected '&' in Standard input code"

sebastienbarre:
@grongor @kukulich Thanks for the quick fix.

Now I wanted to mention something that has been pointed out to me since then.

This is illegal in PHP:
```php
$ref = $condition ? &$foo : &$bar;
```
but this is legal apparently:
```php
$ref = &${$condition ? 'foo' : 'bar'};
```

Your call. Personally I find it pretty ugly, so I'm not sure I would want to force people to do it.

grongor:
oh, man ... yeah, I wouldn't mention it anywhere and just hope that everybody else will eventually forget it's possible :joy:

PR #828, opened 2020-01-15
Fix RequireTernaryOperatorSniff for variable refs

grongor:
Fixes https://github.com/slevomat/coding-standard/issues/826


tests/Sniffs/ControlStructures/data/requireTernaryOperatorNoErrors.php
```diff
@@ -78,3 +78,11 @@
 if (true) {
 	echo 'true';
 }
+
+$foo = 1;
+$bar = 2;
+if (true) {
```
kukulich: It would be better to have two tests:

1. `&` in `if`
2. `&` in `else`

kukulich:
Thanks.

Issue #678, opened 2019-05-05
Feature request: Control Structure Spacing for parent call (mainly constructor)


Levivb:
Hey,

Great work on the previously implemented `SlevomatCodingStandard.ControlStructures.ControlStructureSpacing`.<br>
This request is a bit of a mash-up of both `ControlStructureSpacing` rule and [Feature request: Call parent::__construct before other constructor code ](https://github.com/slevomat/coding-standard/issues/447)

Would it be possible to add support for spacing for parent calls?

Example:
```php
public function __construct(
        WorkAreaRepository $workAreaRepository,
        InternalEducationRepository $internalEducationRepository,
        AgeGroupRepository $ageGroupRepository
    ) {
        parent::__construct();
        $this->workAreaRepository = $workAreaRepository;
        $this->internalEducationRepository = $internalEducationRepository;
        $this->ageGroupRepository = $ageGroupRepository;
    }
```

would become:
```php
public function __construct(
        WorkAreaRepository $workAreaRepository,
        InternalEducationRepository $internalEducationRepository,
        AgeGroupRepository $ageGroupRepository
    ) {
        parent::__construct();

        $this->workAreaRepository = $workAreaRepository;
        $this->internalEducationRepository = $internalEducationRepository;
        $this->ageGroupRepository = $ageGroupRepository;
    }
```

when applying the following:
```
<rule ref="SlevomatCodingStandard.ControlStructures.ControlStructureSpacing/OrOtherRuleName">
        <properties>
            <property name="linesCountAroundControlStructure" value="1" />
            <property name="linesCountBeforeFirstControlStructure" value="0" />
            <property name="linesCountAfterLastControlStructure" value="0" />
        </properties>
    </rule>
```

A parent call doesn't seem a separate token, so I reckon it would be difficult to add support for that in the current `ControlStructureSpacing` rule. So a different rule might be necessary, but you might have other bright ideas

Let me know what you think!

PR #829, opened 2020-01-15
Add parent call spacing sniff

grongor:
Fixes https://github.com/slevomat/coding-standard/issues/678

I'm not sure if we should create a separate sniff extending the abstract class, or just keep it here. What do you think? Splitting is trivial ... just pitch me a good name :D

### DIFF
--- a/tests/Sniffs/Classes/ParentCallSpacingSniffTest.php
+++ b/tests/Sniffs/Classes/ParentCallSpacingSniffTest.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Classes;
+
+use SlevomatCodingStandard\Sniffs\TestCase;
+
+class ParentCallSpacingSniffTest extends TestCase
+{
+
+	public function testNoErrors(): void
+	{
+		$report = self::checkFile(__DIR__ . '/data/parentCallSpacingNoErrors.php');
+		self::assertNoSniffErrorInFile($report);
+	}
+
+	public function testErrors(): void
+	{
+		$report = self::checkFile(__DIR__ . '/data/parentCallSpacingErrors.php');
+
+		self::assertSame(2, $report->getErrorCount());
+
+		self::assertSniffError($report, 5, ParentCallSpacingSniff::CODE_INCORRECT_LINES_COUNT_AFTER_CONTROL_STRUCTURE, 'Expected 1 lines after "parent", found 0.');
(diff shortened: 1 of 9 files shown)


kukulich:
I definitely want new sniff :) It should be in `Classes` directory.

We probably  should support `self::call()` too.

And more tests should be needed for codes like:

```
foreach (self::getAvailableEnums() as $interval) {
```

grongor:
So put it in Classes directory and extend from here?

Also, don't think that the `self::call()` makes much sense, as it's just a generic method call ... if you want to support that, then we should also support regular methods. Yeah, it doesn't make sense to me.

I was a bit on edge about `static::call()` too, but I think it's "special enough" and more like `parent::call()`.

Anyway, what would be the sniff name? :)

kukulich:
I think we can create `Classes.ParentCallSpacingSniff` and support only `parent`.

The `static::call()` is strange case for me. If you want to support it too I would add second sniff `Classes.StaticCallSpacingSniff`

grongor:
It's not very much used, but yeah I consider it a kind of special case...but yeah, I talked with my colleague and we decided that it's not worth to have a sniff for it :D I'll post the parent sniff soon.

tests/Sniffs/Classes/ParentCallSpacingSniffTest.php
```diff
@@ -0,0 +1,29 @@
...
+		self::assertNoSniffErrorInFile($report);
+	}
+
+	public function testErrors(): void
+	{
+		$report = self::checkFile(__DIR__ . '/data/parentCallSpacingErrors.php');
+
+		self::assertSame(2, $report->getErrorCount());
+
+		self::assertSniffError($report, 5, AbstractControlStructureSpacingSniff::CODE_INCORRECT_LINES_COUNT_AFTER_CONTROL_STRUCTURE, 'Expected 1 lines after "parent", found 0.');
```
kukulich: Use `ParentCallSpacingSniff::`

grongor: yeah ofc

grongor:
> 2. I don't know how to disable useless settings `tokensToCheck` for this sniff :(

Maybe I just fill the property with this sniff, so it's there automatically? ... you kind off broke this in here https://github.com/slevomat/coding-standard/commit/307fcc2c418d24d7c776e7681278b8cb2e8d7f39 :D

kukulich:
We need to move the property `$tokensToCheck` to `BlockControlStructureSpacingSniff` and `JumpStatementsSpacingSniff`.

grongor:
Done

Issue #624, opened 2019-03-04
Sort methods by scope


loic-lopez:
Hello, thanks for creating this package it's awesome!

There is a rule to sort methods of a class by their scope for example?

```php
class Example {
   // public scope
   public $foo;
   public $foo2;

   // protected scope
   protected $bar;
   protected $bar2;
}

```

Thanks for reply 👍

andreasschroth:
I like this idea, we also always sort methods and properties by visibility.

In your example it's properties though, not methods.

loic-lopez:
Thanks for reply, and yes my bad i just not insert public/protected/privatemethods in my example but the main idea of sorting methods/properties by their visibilty is great i think.

loic-lopez:
This is not possible actualy with the latest version of coding-standard?

And if not could this issue be transformed to a feature request?

janedbal:
It is implemented here: https://github.com/cdn77/coding-standard/blob/v1.1.1/src/Cdn77/Sniffs/Classes/ClassStructureSniff.php

loic-lopez:
@janedbal thanks for reply but how to include this sniff in ruleset.xml?

Issue #701, opened 2019-05-22
New sniffs proposal


grongor:
We have some custom sniffs in our company's CS and I think more people could benefit from them if we ported them to your repository. I'd be willing to do such migration myself if it meant only slight changes.

So if you are interested, please take a look at our sniffs at https://github.com/cdn77/coding-standard/tree/master/src/Cdn77/Sniffs (ignore ControlStructures - they are just duplicated from your repository).

kukulich:
@grongor It would be superb :) Please open special PR for each sniff :)

I only don't know why you have `LineLengthSniff`. There's sniff in PHPCS itself.

grongor:
Alright, I will when I find some spare time.

I'm not sure either but I know that there was some reason :D I think that it didn't handle all the situations correctly or something like that ... I will look into it.

Same goes for OperatorSpacingSniff, which has multiple implementations and all are flawed in some way. Mine for example also handles unary operators like `T_INC` and `T_DEC`.

kukulich:
@grongor Understood, we have also our own `TrailingArrayCommaSniff` :) We have also better environment for writing tests than PHPCS itself :)

I'll try to release version 5.1. asap, so all of your PR will be part of 5.2.

jrfnl:
> Mine for example also handles unary operators like T_INC and T_DEC.

PHPCS has a native sniff for spacing of unary operators since PHPCS 3.4.0 `Generic.WhiteSpace.IncrementDecrementSpacing`, so this may not be needed anymore.

grongor:
@jrfnl yeah, indeed the `T_INC` and `T_DEC` are handled by this sniff already... but I'm still missing these features:
- handling of unary `T_MINUS`, eg. `$a = - $b` -> `$a = -$b`
- handling of `T_STRING_CONCAT` and `T_INSTANCEOF` (Squiz sniff doesn't allow me to add these types)

Would you propose a better solution then handling all the operators in a single sniff as I'm doing now?

jrfnl:
> handling of unary T_MINUS, eg. $a = - $b -> $a = -$b

The `Squiz.Formatting.OperatorBracket.SpacingAfterMinus` comes close as it triggers on `$b = - 10;`, so I think that sniff would only need a little adjustent to also handle `$a = - $b`.

> handling of T_STRING_CONCAT and T_INSTANCEOF (Squiz sniff doesn't allow me to add these types)

[...]

PR #830, opened 2020-01-15
Add ClassStructureSniff

grongor:
Part of https://github.com/slevomat/coding-standard/issues/701 series.

Fixes https://github.com/slevomat/coding-standard/issues/624

:wrench:  TODO: add readme if sniff is accepted

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/ClassStructureSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/ClassStructureSniff.php
@@ -1,0 +1,459 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use SlevomatCodingStandard\Helpers\ClassHelper;
+use SlevomatCodingStandard\Helpers\DocCommentHelper;
+use SlevomatCodingStandard\Helpers\FunctionHelper;
+use SlevomatCodingStandard\Helpers\TokenHelper;
+use function array_key_exists;
+use function array_merge;
+use function assert;
+use function in_array;
+use function sprintf;
+use function str_repeat;
+use function strtolower;
+use const T_ABSTRACT;
+use const T_CLOSE_CURLY_BRACKET;
+use const T_CONST;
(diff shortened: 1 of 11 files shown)


SlevomatCodingStandard/Sniffs/Classes/ClassStructureSniff.php
```diff
@@ -0,0 +1,478 @@
...
+
+		$file->fixer->endChangeset();
+	}
+
+	private function findPreviousMemberEndPointer(File $file, int $memberPointer): int
+	{
+		$endTypes = [T_OPEN_CURLY_BRACKET, T_CLOSE_CURLY_BRACKET, T_SEMICOLON];
+		$previousMemberEndPointer = TokenHelper::findPrevious($file, $endTypes, $memberPointer - 1);
+		if ($previousMemberEndPointer === null) {
+			throw new RuntimeException('Previous member end pointer not found');
```
kukulich: Are we able to remove all `throw new RuntimeException` ?

grongor: Sure, I'll replace it with asserts.

kukulich:
LGTM.

kukulich:
@grongor How can I set different order in my ruleset? Can you please post an example?

kukulich:
@grongor It looks it does not work at all :) I've already refectored (and fixed) it on my localhost :)

grongor:
Sorry for the late response ... I'm sorry, I never actually tested the settings outside the tests ... I think it worked, just not with the constants. Thanks for the fix.

grongor:
There is a little issue with your changes: you can't set multiple groups to have the same expected order (ergo be ignored by the check). That might've been useful for people who wanted to check "just that constants are before properties" or something like that. Also, if you don't specify some groups in the custom config then they will be expected to be at the end of the class, which doesn't make much sense to me.

I personally don't care about these issues, I just wanted to point out that they now exist.

kukulich:
@grongor You can :) See readme and example https://github.com/slevomat/coding-standard#slevomatcodingstandardclassesclassstructure-

grongor:
Oh nice, I just tried to figure it out from the code and it seemed at the first glance that it's not possible. Perfect then! Missing groups will still be at the end, but I guess that's ok :)

kukulich:
Yeah, I think you can just specify all groups and you're ok :)

PR #831, opened 2020-01-15
Add RequireSingleLineMethodDeclarationSniff

grongor:
Part of #701 series.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,6 +309,14 @@ Sniff provides the following settings:
 * `linesCountAfterControlStructure`: allows to configure the number of lines after parent call.
 * `linesCountAfterLastControlStructure`: allows to configure the number of lines after last parent call.
 
+#### SlevomatCodingStandard.Classes.RequireSingleLineMethodDeclaration 🔧
+
+Enforces method declaration to be on a single line if it fits on it (or all the time; depends on settings).
+
+Sniff provides the following settings:
+
+* `maxLineLength`: specifies max allowed line length; if declaration would fit on it, it's enforced. Use 0 value to enforce for all methods, regardless of length.
+
 #### SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming
 
 Reports use of superfluous prefix or suffix "Abstract" for abstract classes.
--- a/SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodDeclarationSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodDeclarationSniff.php
@@ -1,0 +1,102 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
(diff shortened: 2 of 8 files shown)


SlevomatCodingStandard/Sniffs/Classes/SingleLineMethodDeclarationWhenPossibleSniff.php
```diff
@@ -0,0 +1,104 @@
...
+use function rtrim;
+use function sprintf;
+use function str_replace;
+use function strlen;
+use const T_FUNCTION;
+use const T_OPEN_CURLY_BRACKET;
+use const T_SEMICOLON;
+use const T_WHITESPACE;
+
+class SingleLineMethodDeclarationWhenPossibleSniff implements Sniff
```
kukulich: What about RequireSingleLineMethodDeclarationSniff? I don't like this name :)

SlevomatCodingStandard/Sniffs/Classes/SingleLineMethodDeclarationWhenPossibleSniff.php
```diff
@@ -0,0 +1,104 @@
...
+{
+
+	public const CODE_UNNECESSARY_MULTI_LINE_METHOD = 'UnnecessaryMultiLineMethod';
+
+	/** @var int */
+	public $maxLineLength = 120;
```
kukulich: Will `0` work for all methods?

SlevomatCodingStandard/Sniffs/Classes/SingleLineMethodDeclarationWhenPossibleSniff.php
grongor: Sure, makes sense :) I didn't really think about it when copying it over here :D

SlevomatCodingStandard/Sniffs/Classes/SingleLineMethodDeclarationWhenPossibleSniff.php
grongor: Not sure if I understand, but ... if you set this to 0, then this sniff effectively does nothing, because all methods will be considered as not inlinable. If you set this to lets say 10000, all methods will be required to have an inline definition.

kukulich: @grongor Ok, can we support `null` (or probably empty string because of PHPCS configuration) to require all methods to be on one line?

grongor: Ok, what if I adjusted the code to use 0 value for that? Because having it "do nothing" doesn't make sense and it's just a side-effect.

kukulich: Ok :)

kukulich:
LGTM.

PR #832, opened 2020-01-15
Add PropertySpacingSniff and ConstantSpacingSniff

grongor:
Part of #701 series.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
@@ -1,0 +1,102 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\TokenHelper;
+use function assert;
+use function in_array;
+use const T_COMMENT;
+use const T_CONST;
+use const T_DOC_COMMENT_OPEN_TAG;
+use const T_FUNCTION;
+use const T_PRIVATE;
+use const T_PROTECTED;
+use const T_PUBLIC;
+use const T_SEMICOLON;
+use const T_STRING;
+use const T_VAR;
+use const T_VARIABLE;
+
+abstract class AbstractPropertyAndConstantSpacingSniff implements Sniff
+{
+
+	/** @var int */
+	public $minLinesCountBeforeWithComment = 1;
+
+	/** @var int */
+	public $maxLinesCountBeforeWithComment = 1;
+
+	/** @var int */
+	public $minLinesCountBeforeWithoutComment = 0;
+
+	/** @var int */
+	public $maxLinesCountBeforeWithoutComment = 1;
+
+	/**
+	 * @phpcsSuppress SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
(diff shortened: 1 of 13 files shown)


kukulich:
I think it would be better to split this sniff to `PropertySpacingSniff` and `ConstantSpacingSniff`. And remove "Property" string from all options.

grongor:
Done

SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
```diff
@@ -0,0 +1,102 @@
...
+use const T_SEMICOLON;
+use const T_STRING;
+use const T_VAR;
+use const T_VARIABLE;
+
+abstract class AbstractPropertyAndConstantSpacingSniff implements Sniff
+{
+
+	/** @var int */
+	public $minLinesCountBeforeWithComment = 1;
```
kukulich: It think it would be better check exactly the count of lines. Why is `min` here?

SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/** @var int */
+	public $maxLinesCountBeforeWithComment = 1;
+
+	/** @var int */
+	public $minLinesCountBeforeWithoutComment = 0;
```
kukulich: Please set the default value to `1`.

SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
grongor: This provides you with more flexibility. If you want to force exact number of lines, just set both min and max to the same number. Let's say someone likes to stack the properties, so they'll set min to 0. But they would also like to add an empty line in between properties to visually separate let's say, two groups. To do that, they set max to 1. You wouldn't be able to do that with just one value. It would be either stacking or space.

SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
grongor: Sure, makes sense.

grongor: You sure? It doesn't look that good to me ... and seems unnecessary.

```php
final class OperationalStatus extends Enum
{
    public const UP = 1;
    public const DOWN = 2;
    public const TESTING = 3;
    public const UNKNOWN = 4;
    public const DORMANT = 5;
    public const NOT_PRESENT = 6;
    public const LOWER_LAYER_DOWN = 7;
...
```
vs
```php
final class OperationalStatus extends Enum
{
    public const UP = 1;

    public const DOWN = 2;

    public const TESTING = 3;

    public const UNKNOWN = 4;

    public const DORMANT = 5;

    public const NOT_PRESENT = 6;

    public const LOWER_LAYER_DOWN = 7;

...
```

kukulich: You're right.

SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacingSniff.php
kukulich: Ok.

kukulich:
LGTM.

PR #835, opened 2020-01-15
Add NegationOperatorSpacingSniff

grongor:
Part of #701 series.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Operators/NegationOperatorSpacingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Operators/NegationOperatorSpacingSniff.php
@@ -1,0 +1,94 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Operators;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\TokenHelper;
+use function in_array;
+use function strlen;
+use const T_CLOSE_PARENTHESIS;
+use const T_CLOSE_SHORT_ARRAY;
+use const T_CLOSE_SQUARE_BRACKET;
+use const T_CONSTANT_ENCAPSED_STRING;
+use const T_DNUMBER;
+use const T_ENCAPSED_AND_WHITESPACE;
+use const T_LNUMBER;
+use const T_MINUS;
+use const T_NUM_STRING;
+use const T_STRING;
+use const T_VARIABLE;
+use const T_WHITESPACE;
+
+final class NegationOperatorSpacingSniff implements Sniff
+{
+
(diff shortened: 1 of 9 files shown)


SlevomatCodingStandard/Sniffs/Operators/NegationOperatorSpacingSniff.php
```diff
@@ -0,0 +1,99 @@
...
+use const T_VARIABLE;
+use const T_WHITESPACE;
+
+final class NegationOperatorSpacingSniff implements Sniff
+{
+
+	public const CODE_INVALID_SPACE_AFTER_MINUS = 'InvalidSpaceAfterMinus';
+
+	/** @var bool */
+	public $requireSpace = false;
```
kukulich: Please change to integer to be consistent with other sniffs.

tests/Sniffs/Operators/NegationOperatorSpacingSniffTest.php
```diff
@@ -0,0 +1,27 @@
...
+	public function testNoErrors(): void
+	{
+		self::assertNoSniffErrorInFile(self::checkFile(__DIR__ . '/data/NegationOperatorSpacingSniffNoErrors.php'));
+	}
+
+	public function testErrors(): void
+	{
+		$file = self::checkFile(
+			__DIR__ . '/data/NegationOperatorSpacingSniffRequireSpaceErrors.php',
+			['requireSpace' => true]
```
kukulich: Please add both tests (errors and no errors) also for `false` (or `0` after the change).

SlevomatCodingStandard/Sniffs/Operators/NegationOperatorSpacingSniff.php
grongor: You mean to change it to "requiredSpacesCount"? You sure that makes sense? Or what other sniff uses integer instead of boolean, and why? Doesn't make sense to me.

tests/Sniffs/Operators/NegationOperatorSpacingSniffTest.php
grongor: There is no-errors for withoutSpace variant, and with errors for with space variant ... isn't that enough?

kukulich: Is not. Now you don't test fixer for `requireSpace=false`

SlevomatCodingStandard/Sniffs/Operators/NegationOperatorSpacingSniff.php
kukulich: Yes, I mean `$spacesCount`. I use always integer so I want to be consistent with other sniffs.

tests/Sniffs/Operators/NegationOperatorSpacingSniffTest.php
grongor: Yeah makes sense.

kukulich:
LGTM.

PR #837, opened 2020-01-15
Add LineLengthSniff

grongor:
Part of #701 series.

As discussed, this sniff is kind of duplicate of one in Squiz, but this one handles some exceptions.


kukulich:
Just for my info: why is this sniff better?

It looks good so please update README.

grongor:
Removed unnecessary things like metrics, warnings, ... and added an option to ignore use statements.

kukulich:
Ok, so ready the merge when README is added.

PR #840, opened 2020-01-16
EarlyExit: don't break the complete PHPCS run for unsupported syntaxes

jrfnl:
While it is your prerogative, of course, to choose not to support certain syntaxes, when these are encountered the sniff should exit silently.

As things were, the sniff would exit with an _uncaught exception_ which within PHPCS was then translated to an `Internal.Exception` error which stops the PHPCS dead for the file being scanned, this includes breaking the run for all other sniffs in a project ruleset.

Exceptions are intended for developers, not for end-users.

The end-user of PHPCS can do nothing with an uncaught exception. Instead, this exception should be caught within the sniff and the sniff should return silently as it couldn't be determined whether something is or isn't an error for the rules the sniff is checking for.

Includes adjusted unit tests + a new unit test for control structures using alternative syntax, which clearly isn't supported either and would - until now - cause an `Undefined index: scope_condition` notice, which - again - would stop the PHPCS scan dead.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/ControlStructures/EarlyExitSniff.php
+++ b/SlevomatCodingStandard/Sniffs/ControlStructures/EarlyExitSniff.php
@@ -8,6 +8,7 @@ use PHP_CodeSniffer\Sniffs\Sniff;
 use SlevomatCodingStandard\Helpers\ConditionHelper;
 use SlevomatCodingStandard\Helpers\IndentationHelper;
 use SlevomatCodingStandard\Helpers\TokenHelper;
+use Throwable;
 use function array_key_exists;
 use function in_array;
 use function sort;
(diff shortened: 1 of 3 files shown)


kukulich:
@jrfnl Thanks, looks good!

jrfnl:
@kukulich Thanks ;-)

The same principle should probably be applied in a few more places.... I've run into this with other Slevomat sniffs too at times (just didn't get annoyed enough yet to fix it before).

kukulich:
@jrfnl Please open issue as a reminder to fix the rest and we will see who will have time to fix it :)

jrfnl:
@kukulich Done. See #841

kukulich:
@jrfnl Fix build please.

jrfnl:
@kukulich Done :crossed_fingers: The other PRs will probably need this too. I tried to run PHPCS locally on the PRs, but was getting a `Uncaught Error: Class 'PHPStan\PhpDocParser\Parser\TokenIterator' not found`, so I'll just have to rely on Travis in this case.

Efficiency suggestion: maybe move the coding standards run up in the stages ? No need to run the more expensive unit tests when the build would fail on CS anyhow.

Issue #841, opened 2020-01-16
Don't exit sniffs with an exception, return silently instead


jrfnl:
From #840 - the same principle as applied in that PR should also be applied to other sniffs:

> While it is your prerogative, of course, to choose not to support certain syntaxes, when these are encountered the sniff should exit silently.
>
> As things were, the sniff would exit with an uncaught exception which within PHPCS was then translated to an Internal.Exception error which stops the PHPCS dead for the file being scanned, this includes breaking the run for all other sniffs in a project ruleset.
> 
> Exceptions are intended for developers, not for end-users.
>
> The end-user of PHPCS can do nothing with an uncaught exception. Instead, this exception should be caught within the sniff and the sniff should return silently as it couldn't be determined whether something is or isn't an error for the rules the sniff is checking for.

jrfnl:
I've just done a quick search on the unit test files to see where expected exceptions were being tested.

That yielded three sniffs. PRs to fix these have been submitted (#842, #843, #844).

No guarantee that this covers all situations where this is happening as I don't know whether all sniffs which throw exceptions also have unit tests covering that situation.

kukulich:
@jrfnl All sniffs has 100 % code coverage :)

jrfnl:
@kukulich If you can, you may want to stop some of the older builds on Travis as I've repushed all branches one or more times (and possibly turn the auto-cancellation options on).

jrfnl:
As all above mentioned PRs have now been merged, I'm closing this issue. I may reopen if I run into more of the same, but hopefully there won't be any, what with the 100% code coverage :grinning:

PR #842, opened 2020-01-16
UselessIfConditionWithReturn: don't break the complete PHPCS run for unsupported syntaxes

jrfnl:
While it is your prerogative, of course, to choose not to support certain syntaxes, when these are encountered the sniff should exit silently.

As things were, the sniff would exit with an _uncaught exception_ which within PHPCS was then translated to an `Internal.Exception` error which stops the PHPCS dead for the file being scanned, this includes breaking the run for all other sniffs in a project ruleset.

Exceptions are intended for developers, not for end-users.

The end-user of PHPCS can do nothing with an uncaught exception. Instead, this exception should be caught within the sniff and he sniff should return silently as it couldn't be determined whether something is or isn't an error for the rules the sniff is checking for.

Includes adjusted unit tests.

Related to #841


kukulich:
LGTM.

PR #845, opened 2020-01-16
Travis: fail the build early if CS/PHPStan would fail

jrfnl:
... instead of running the (more expensive) unit tests first.

This will give quicker build results for builds which would fail on CS/PHPStan.


kukulich:
I'm sorry I did not read this and do it by myself in master :)

jrfnl:
@kukulich No worries, at least it's done ;-)

PR #846, opened 2020-01-16
Bug fix: rename AbstractControlStructureSpacingSniff (remove Sniff suffix)

jrfnl:
PHPCS automatically loads all files with a `Sniff.php` suffix in `Sniffs/Category` directories of a standard, presuming they are actual sniffs.
And if the complete standard is included, it will (generally speaking) do so in category -> file alphabetic order.

As - with that logic - the newly introduced `ParentCallSpacingSniff` is loaded before the sniffs in the `ControlStructures` category and extends the `AbstractControlStructureSpacingSniff` which will therefore be autoloaded, a `Fatal error:  Cannot declare class SlevomatCodingStandard\Sniffs\ControlStructures\AbstractControlStructureSpacingSniff, because the name is already in use`  will be the result when the PHPCS autoloader reaches the `ControlStructures` folder.

The PHPCS autoloader will also handle the loading of non-Sniff files in the `SlevomatCodingStandard` directory and subdirectories without problem or conflict, they just shouldn't be suffixed with `Sniff`.

Having said all that, this PR renames the `AbstractControlStructureSpacingSniff` class to `AbstractControlStructureSpacing`, including renaming the file.

This solves the conflict.

Alternatively, it could be elected to move all `abstract` base sniff classes to, for instance, an `Abstracts` subdirectory of `SlevomatCodingStandard`. That should also solve the problem.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacing.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacing.php
@@ -19,7 +19,7 @@ use const T_STRING;
 use const T_VAR;
 use const T_VARIABLE;
 
-abstract class AbstractPropertyAndConstantSpacingSniff implements Sniff
+abstract class AbstractPropertyAndConstantSpacing implements Sniff
 {
 
 	/** @var int */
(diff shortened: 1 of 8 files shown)


jrfnl:
I've added a second commit which fixes the same (potential) issue for the `AbstractPropertyAndConstantSpacingSniff` as newly introduced in #832.

kukulich:
LGTM.

kukulich:
Thanks.

PR #854, opened 2020-01-21
MethodSignatureSniffs: Add included and exluded patters

RiKap:
Hey.
I added `includedMethodPatterns` and `excludedMethodPatterns` property for `RequireSingleLineMethodSignatureSniff` and `RequireMultiLineMethodSignatureSniff`

**includedMethodPatterns**:
It will allow you to set methods which you would like to check by these two sniffs

**excludedMethodPatterns**:
It will allow you to set methods which you would like to ignore by these two sniffs

--------

Sometimes it is good to enforce `RequireMultiLineMethodSignatureSniff` just for one method, for example `__construct`.

Another benefit of this change is that you are able to use these two sniffs together.
It will force `RequireMultiLineMethodSignature` for `__construct` and `RequireSignleLineMethodSignature` for all other methods.
```
<rule ref="SlevomatCodingStandard.Classes.RequireMultiLineMethodSignature">
	<properties>
		<property name="minLineLength" type="int" value="0"/>
		<property name="includedMethodPatterns" type="array" value="
			/__construct(.*)/,
		"/>
	</properties>
</rule>
<rule ref="SlevomatCodingStandard.Classes.RequireSignleLineMethodSignature">
	<properties>
		<property name="minLineLength" type="int" value="0"/>
		<property name="excludedMethodPatterns" type="array" value="
			/__construct(.*)/,
		"/>
	</properties>
</rule>
```

There is no BC break. If you don't set `excludedMethodPatterns` and `includedMethodPatterns` it will works as before.


SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
```diff
@@ -60,4 +73,26 @@ public function process(File $phpcsFile, $methodPointer): void
...
 	}
 
+	/**
+	 * @param string $methodSignature
+	 * @param string[] $patterns
+	 * @return bool
+	 */
+	private function isMethodSignatureInPatterns(string $methodSignature, array $patterns): bool
+	{
+		foreach (SniffSettingsHelper::normalizeArray($patterns) as $pattern) {
```
kukulich: Please use `normalizedPatters` so the method `SniffSettingsHelper::normalizeArray()` is not called everytime.

See Eg. https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/Commenting/UselessFunctionDocCommentSniff.php#L130

SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
```diff
@@ -38,6 +44,13 @@ public function process(File $phpcsFile, $methodPointer): void
 		$signature = $this->getSignature($phpcsFile, $signatureStartPointer, $signatureEndPointer);
 		$signatureWithoutTabIndetation = $this->getSignatureWithoutTabs($signature);
 
+		if (count($this->includedMethodPatterns) !== 0) {
+			$isMethodSignatureInPatterns = $this->isMethodSignatureInPatterns($signatureWithoutTabIndetation, $this->includedMethodPatterns);
```
kukulich: Use `FunctionHelper::getName()` and check only method names with patterns.

And please remove useless variable `$isMethodSignatureInPatterns` and join both `if` together.

README.md
```diff
@@ -392,6 +392,8 @@ Sniff provides the following settings:
 
 * `includedMethodPatterns`: allows to configure which methods are included in sniff detection. This is an array of regular expressions (PCRE) with delimiters.
 
+* `excludedMethodPatterns`: allows to configure which methods are excluded from sniff detection. This is an array of regular expressions (PCRE) with delimiters.
+
```
kukulich: Please document how it works when you set both options together.

SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
```diff
@@ -60,4 +73,26 @@ public function process(File $phpcsFile, $methodPointer): void
...
+	 * @return bool
+	 */
+	private function isMethodSignatureInPatterns(string $methodSignature, array $patterns): bool
+	{
+		foreach (SniffSettingsHelper::normalizeArray($patterns) as $pattern) {
+			if (!SniffSettingsHelper::isValidRegularExpression($pattern)) {
+				throw new Exception(sprintf('%s is not valid PCRE pattern.', $pattern));
+			}
+
+			if (preg_match($pattern, $methodSignature) === 0) {
```
kukulich: Use negative condition and `return true` please.

SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
```diff
@@ -60,4 +73,26 @@ public function process(File $phpcsFile, $methodPointer): void
...
+	/**
+	 * @param string $methodSignature
+	 * @param string[] $patterns
+	 * @return bool
+	 */
+	private function isMethodSignatureInPatterns(string $methodSignature, array $patterns): bool
+	{
+		foreach (SniffSettingsHelper::normalizeArray($patterns) as $pattern) {
+			if (!SniffSettingsHelper::isValidRegularExpression($pattern)) {
+				throw new Exception(sprintf('%s is not valid PCRE pattern.', $pattern));
```
kukulich: Please add unit test to test this line.

See `ForbiddenCommentsSniffTest`

RiKap:
@kukulich 
2. Why it does not make sense now, but it did before?

kukulich:
Now you check only method names.

kukulich:
Thanks.

PR #861, opened 2020-01-23
Fix RequireSingleLineMethodSignature for semicolons

grongor:
Not sure why would you @kukulich change the test when it was working fine before ... anyway, I reverted your change and added a fix.

Without this, the phpcs was jumping between LineLengthSniff error when on single line and with error here when on multiple lines...

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/AbstractMethodSignature.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/AbstractMethodSignature.php
@@ -46,6 +46,9 @@ abstract class AbstractMethodSignature implements Sniff
 		assert(!is_bool($signatureStartPointer));
 
 		$pointerAfterSignatureEnd = TokenHelper::findNext($phpcsFile, [T_OPEN_CURLY_BRACKET, T_SEMICOLON], $methodPointer + 1);
+		if ($phpcsFile->getTokens()[$pointerAfterSignatureEnd]['code'] === T_SEMICOLON) {
+			return [$signatureStartPointer, $pointerAfterSignatureEnd];
+		}
 
 		$signatureEndPointer = TokenHelper::findPreviousEffective($phpcsFile, $pointerAfterSignatureEnd - 1);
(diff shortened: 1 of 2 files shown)


SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
```diff
@@ -36,10 +36,10 @@ public function process(File $phpcsFile, $methodPointer): void
 		}
 
 		$signature = $this->getSignature($phpcsFile, $signatureStartPointer, $signatureEndPointer);
```
kukulich: Please revert all changes in this file.

kukulich:
I'm not sure why I changed it. I tested it on whole Slevomat and there was some bugs I've fixed etc.

SlevomatCodingStandard/Sniffs/Classes/RequireSingleLineMethodSignatureSniff.php
grongor: Why? There is a typo + "double normalization", both should be addressed.

kukulich: Because it's not relevant to the bug. And your typo fix is not right :) "Indention" is wrong too :)

https://translate.google.cz/#view=home&op=translate&sl=en&tl=cs&text=indentation

grongor: I just used the first one that Phpstorm offered. According to the dictionary it's correct, but archaic ... alright, so fix in spearate commit or you wan't completely separate pr?

kukulich: Just ignore it :) I will fix it by myself but I'm happy you've found it :)

PR #862, opened 2020-01-23
Allow suppress for NullableTypeForNullDefaultValue sniff

simPod:
Eg. Symfony still supports ancient PHP 7.1 so type hints are often declared as

```php
    /**
     * Gets the listeners of a specific event or all listeners sorted by descending priority.
     *
     * @return array The event listeners for the specified event, or all event listeners by event name
     */
    public function getListeners(string $eventName = null);
```
 with missing `?` for nullable types. It is not possible to type hint implementations as 

```php
    public function getListeners(?string $eventName = null) {}
```

Therefore there's need for suppressing the `SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue.NullabilitySymbolRequired`

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,7 +735,7 @@ Sniff provides the following settings:
 
 * `spacesCountBeforeColon`: the number of spaces expected between closing brace and colon.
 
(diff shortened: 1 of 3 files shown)


kukulich:
Thanks.

Issue #866, opened 2020-01-24
Additional setting for TraitUseSpacing


thled:
Sniff: SlevomatCodingStandard.Classes.TraitUseSpacing
Setting: linesCountBeforeFirstUseWhenFirstInClass: allows to configure the number of lines before first use when the use is the first statement in the class.

There is already the similar sniff "linesCountAfterLastUseWhenLastInClass", but for me it is way more common to declare the used traits at the beginning of the class. And I do not want to begin the class with an empty line.

It is also recommended by [PSR12 Rule 4.2](https://www.php-fig.org/psr/psr-12/#42-using-traits).

kukulich:
@thled What is your request? I really don't understand.

thled:
I would like to add a setting to the Sniff "SlevomatCodingStandard.Classes.TraitUseSpacing" which allows me to configure the number of lines before the first "use" when that "use" is the first statement in a class.

Example:

```php
final class foo
{
    // x empty lines
    use BarTrait;
    // y empty lines
}
```

We already can configure the "y" ("linesCountAfterLastUseWhenLastInClass") but not the "x" (I propose "linesCountBeforeFirstUseWhenFirstInClass").

kukulich:
You can set it with `EmptyLinesAroundClassBracesSniff`

thled:
If that would be the solution, than why is the "linesCountAfterLastUseWhenLastInClass" setting there? It is effected the same way by your proposed sniff.

Another reason is I should not rely on that other sniff if I just want "use" for traits to be a set number lines after the opening brace of a class.

It feels consistent to offer my proposed setting, but it is Ok if you do not want to add it. I will find ways to cover my use case.

kukulich:
@thled Will you prepare PR?

thled:
I will try it next week. Thank you for your quick responses.

PR #870, opened 2020-01-28
TraitUseSpacingSniff: Add linesCountBeforeFirstUseWhenFirstInClass #866

thled:
Hello,

I prepared this PR for my issue #866 from last week.

It adds a `linesCountBeforeFirstUseWhenFirstInClass` setting to the `SlevomatCodingStandard.Classes.TraitUseSpacing` sniff.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/TraitUseSpacingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/TraitUseSpacingSniff.php
@@ -29,6 +29,9 @@ class TraitUseSpacingSniff implements Sniff
 	public $linesCountBeforeFirstUse = 1;
 
 	/** @var int */
+	public $linesCountBeforeFirstUseWhenFirstInClass = null;
+
+	/** @var int */
 	public $linesCountBetweenUses = 0;
 
 	/** @var int */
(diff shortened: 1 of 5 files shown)


SlevomatCodingStandard/Sniffs/Classes/TraitUseSpacingSniff.php
```diff
@@ -28,6 +28,9 @@ class TraitUseSpacingSniff implements Sniff
 	/** @var int */
 	public $linesCountBeforeFirstUse = 1;
 
+	/** @var int */
+	public $linesCountBeforeFirstUseWhenFirstInClass = 1;
```
kukulich: Please change the default to `null` to be compatible with actual version. If the variable is `null`, use `linesCountBeforeFirstUse`.

SlevomatCodingStandard/Sniffs/Classes/TraitUseSpacingSniff.php
```diff
@@ -69,16 +72,19 @@ public function process(File $phpcsFile, $classPointer): void
 
 	private function checkLinesBeforeFirstUse(File $phpcsFile, int $firstUsePointer): void
 	{
+		$tokens = $phpcsFile->getTokens();
+
 		/** @var int $pointerBeforeFirstUse */
 		$pointerBeforeFirstUse = TokenHelper::findPreviousExcluding($phpcsFile, T_WHITESPACE, $firstUsePointer - 1);
+		$isAtTheStartOfClass = $tokens[$pointerBeforeFirstUse]['code'] === T_OPEN_CURLY_BRACKET;
```
kukulich: It does not work for code like this:

```php
Class Anything 
{
	
	public function whatever()
	{
	}

	use SomeTrait;
}
```

thled: I cannot reproduce an error with your example.
If I do not change the default settings of the sniff, it just throws the error that a blank line is expected after the `use`. Which is correct.
What exactly is not working in your example?

SlevomatCodingStandard/Sniffs/Classes/TraitUseSpacingSniff.php
thled: Done

PR #871, opened 2020-01-28
Fix issue with custom values and int cast

paxal:
When specifying custom value, it's not casted as a string.
Thus we ensure we compare integers together with other integers and fix issue with bad addError() call with string param instead of int.

Fixed error : 
`
PHP Fatal error:  Uncaught TypeError: Argument 4 passed to SlevomatCodingStandard\Sniffs\Classes\PropertySpacingSniff::addError() must be of the type int, string given, called in vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/Classes/AbstractPropertyAndConstantSpacing.php on line 83 and defined in vendor/slevomat/coding-standard/SlevomatCodingStandard/Sniffs/Classes/PropertySpacingSniff.php:25
`


kukulich:
Thanks.

paxal:
Thanks, done

PR #878, opened 2020-01-30
Don't crash on shorthand nullable return types

weirdan:
Previously slevomat/coding-standard would crash (report an Internal.Exception issue) on shorthand nullable return types in a docblock (e.g. `@return ?int`). This change fixes the crash.

Example of a crash
```
FILE: /home/weirdan/src/doctrine-psalm-plugin/stubs/ObjectRepository.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 1 | ERROR | An error occurred during processing; checking has been aborted. The error message was: Undefined property: PHPStan\PhpDocParser\Ast\Type\NullableTypeNode::$types in
   |       | /home/weirdan/src/doctrine-psalm-plugin/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationTypeHelper.php on line 306 (Internal.Exception)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/SlevomatCodingStandard/Helpers/AnnotationTypeHelper.php
+++ b/SlevomatCodingStandard/Helpers/AnnotationTypeHelper.php
@@ -298,11 +298,15 @@ class AnnotationTypeHelper
 	}
 
 	/**
-	 * @param UnionTypeNode|IntersectionTypeNode $typeNode
+	 * @param UnionTypeNode|IntersectionTypeNode|NullableTypeNode $typeNode
 	 * @return bool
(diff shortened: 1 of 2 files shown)


kukulich:
Thank you for your report.

Fixed in master.

PR #895, opened 2020-02-05
Improve class structure sniff

grongor:
Example of new error message:
`The placement of "protected static abstract methods" group is invalid. Last group was "public methods" and one of these is expected after it: public methods, public static methods, protected methods, protected static methods, private methods, private static methods, magic methods`


grongor:
... lol, error in your cs config :D It's already proving useful :)

grongor:
Yeah sorry, rebase error ... fixed

Issue #929, opened 2020-03-22
Namespaces.FullyQualifiedExceptions and Namespaces.UnusedUses causing slow checks


ocean90:
For some yet unknown reasons I noticed that the rules `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions` and `SlevomatCodingStandard.Namespaces.UnusedUses` are causing a much slower check.

In numbers, for the same file (~900 lines): 
* 94ms without any rules
* ~1.3 secs with only `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions`
* ~4 secs with only `SlevomatCodingStandard.Namespaces.UnusedUses`
* ~5.3 secs with `SlevomatCodingStandard.Namespaces.UnusedUses` and `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions`

I can't tell if there is a bug in the sniff or if it's just caused by the design of the sniff to detect the errors and the amount of code in a file. For this specific file [`$referencedNames`](https://github.com/slevomat/coding-standard/blob/e27ab09cb23be30388b7c734353f6218a197e49b/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php#L87) has 211 elements.

I'm happy to share the example file by email since it's private code.

PHP_CodeSniffer version: 3.5.4
PHP version: 7.3.16
Slevomat Coding Standard version: 6.1.5

kukulich:
I dont see anything wrong. Some sniffs are slower than others.

nightlinus:
@ocean90 can u check against master?

ocean90:
@nightlinus Thanks, speed is perfect now! I only noticed that the same error is now printed multiple times, to be exact: 60x. Looks like this:

```
8 | ERROR | [x] Type Something\FOO is not used in this file. (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
8 | ERROR | [x] Type Something\FOO is not used in this file. (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
8 | ERROR | [x] Type Something\FOO is not used in this file. (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
8 | ERROR | [x] Type Something\FOO is not used in this file. (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
[...]

PR #932, opened 2020-03-24
UnusedUsesSniff: improve performance

nightlinus:
As of my test file 8k lines 1 namespace, check was going to be 5s.
With batch namespace lookup check go down to 1s.

Although i don't like that we make extra iterations by referencedNames, but it should be faster without findPrevious() calls.

Relates to #929


kukulich:
Thanks.

kukulich:
I've tried another improvement: https://github.com/slevomat/coding-standard/commit/9dac2807b68b7ff3345c0ce81d1c9bca3df17a5c

PR #935, opened 2020-03-24
UnusedUsesSniff: improve performance in search annotation

nightlinus:
As we already have computed T_NAMESPACE pointers, we should use them in searchAnnotation scenario. Applied same optimization as in 9dac280

On my test file 2x (600ms vs 1200ms) speed improvement.


SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php
```diff
@@ -91,24 +91,17 @@ public function process(File $phpcsFile, $openTagPointer): void
 		if ($pointersBeforeUseStatements === []) {
```
kukulich: If you remove this `if`, it should probably fix the code coverage.

PR #940, opened 2020-03-25
MethodSpacingSniff: add class lookup cache to improve performance

nightlinus:
Increase performance of MethodSpacingSniff by introducing class lookup cache in ClassHelper.

For 8kloc test file time dropped from 2s to 0.4s


kukulich:
@nightlinus Nice, you just have to figure out some workaround for 100 % code coverage :)

PR #948, opened 2020-03-26
Add Whitelist for Full Qualified Namespace Sniffs

bbatsche:
This feature would allow users to select which functions and constants they would like to always be fully qualified, while ignoring others. This is particularly useful for functions where [certain built in functions](https://github.com/php/php-src/blob/f74e30c07c2a94921fbfb7b8936324707505bd75/Zend/zend_compile.c#L3962) may be greatly optimized by making sure they are fully qualified


kukulich:
Thanks.

PR #957, opened 2020-03-31
Remove the basepath from the filename first.

madslundholmdk:
This will ensure to extract the correct expected type name.

---

I had a Laravel project inside a Docker container, and place the project in the folder `app`.
But since Laravel's source folder for the application, is also `app`, since gave me the path `/app/app`.

The sniff `TypeNameMatchesFileNameSniff` couldn't parse the correct expected type name, since both directories is called `app`.
So for the `App` namespace it was: `App\app\...` and for the `Tests` namespace: `Tests\tests\...`

I tried to add `app` to `skipDirs`, but since that is parsed after extracting the root namespace, it only helped on the `app` folder and not on the `tests` folder.

First I just changed the keys for the paths in `rootNamespaces` to be `app/app` and `app/tests`, but I don't think that's very pretty.

Therefor I looked at the source, and added this check on the basepath, and removed it before extrating the expected type name.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Files/TypeNameMatchesFileNameSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Files/TypeNameMatchesFileNameSniff.php
@@ -88,9 +88,12 @@ class TypeNameMatchesFileNameSniff implements Sniff
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


kukulich:
Closed because no response.

PR #958, opened 2020-03-31
AssignmentInConditionSniff: add option ignoreAssignmentsInsideFunctionCalls

nightlinus:
PR adds option to allow such assignments inside calls:

```php
if (in_array(1, $haystack, $strict = true) {

}
```

### DIFF
--- a/SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
+++ b/SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
@@ -5,17 +5,23 @@ namespace SlevomatCodingStandard\Sniffs\ControlStructures;
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
 use SlevomatCodingStandard\Helpers\TokenHelper;
+use function array_keys;
+use function max;
 use function sprintf;
 use const T_DO;
 use const T_ELSEIF;
 use const T_EQUAL;
 use const T_IF;
+use const T_STRING;
 use const T_WHILE;
 
 class AssignmentInConditionSniff implements Sniff
 {
 
 	public const CODE_ASSIGNMENT_IN_CONDITION = 'AssignmentInCondition';
+
+	/** @var bool */
+	public $ignoreAssignmentsInsideFunctionCalls = false;
 
 	/**
 	 * @return array<int, (int|string)>
@@ -54,11 +60,37 @@ class AssignmentInConditionSniff implements Sniff
 
 	private function processCondition(File $phpcsFile, int $parenthesisOpener, int $parenthesisCloser, string $conditionType): void
 	{
-		$equalsTokenPointer = TokenHelper::findNext($phpcsFile, T_EQUAL, $parenthesisOpener + 1, $parenthesisCloser);
-		if ($equalsTokenPointer === null) {
+		$equalsTokenPointers = TokenHelper::findNextAll($phpcsFile, T_EQUAL, $parenthesisOpener + 1, $parenthesisCloser);
+		$tokens = $phpcsFile->getTokens();
+		if ($equalsTokenPointers === []) {
 			return;
 		}
 
+		if (!$this->ignoreAssignmentsInsideFunctionCalls) {
+			$this->error($phpcsFile, $conditionType, $equalsTokenPointers[0]);
+			return;
+		}
+
+		foreach ($equalsTokenPointers as $equalsTokenPointer) {
+			$parenthesisStarts = array_keys($tokens[$equalsTokenPointer]['nested_parenthesis']);
+
+			$insideParenthesis = max($parenthesisStarts);
+			if ($insideParenthesis === $parenthesisOpener) {
+				$this->error($phpcsFile, $conditionType, $equalsTokenPointer);
+				continue;
+			}
+
+			$functionCall = TokenHelper::findPrevious($phpcsFile, T_STRING, $insideParenthesis, $parenthesisOpener);
+			if ($functionCall !== null) {
+				continue;
+			}
+
+			$this->error($phpcsFile, $conditionType, $equalsTokenPointer);
+		}
+	}
+
+	private function error(File $phpcsFile, string $conditionType, int $equalsTokenPointer): void
+	{
 		$phpcsFile->addError(
 			sprintf('Assignment in %s condition is not allowed.', $conditionType),
 			$equalsTokenPointer,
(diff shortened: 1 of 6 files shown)


SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
```diff
@@ -54,11 +62,36 @@ public function process(File $phpcsFile, $conditionStartPointer): void
 
 	private function processCondition(File $phpcsFile, int $parenthesisOpener, int $parenthesisCloser, string $conditionType): void
 	{
-		$equalsTokenPointer = TokenHelper::findNext($phpcsFile, T_EQUAL, $parenthesisOpener + 1, $parenthesisCloser);
-		if ($equalsTokenPointer === null) {
+		$equalsTokenPointers = TokenHelper::findNextAll($phpcsFile, T_EQUAL, $parenthesisOpener + 1, $parenthesisCloser);
+		$tokens = $phpcsFile->getTokens();
+
+		if (!$this->ignoreAssignmentsInsideFunctionCalls && count($equalsTokenPointers) !== 0) {
```
kukulich: Please add `if ($equalsTokenPointers === [])` as first check.

SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
```diff
@@ -54,11 +62,36 @@ public function process(File $phpcsFile, $conditionStartPointer): void
...
+			$this->error($phpcsFile, $conditionType, reset($equalsTokenPointers));
```
kukulich: Don't use `reset`, use `$equalsTokenPointers[0]`

SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
```diff
@@ -54,11 +62,36 @@ public function process(File $phpcsFile, $conditionStartPointer): void
...
 			return;
 		}
 
+		foreach ($equalsTokenPointers as $equalsTokenPointer) {
+			$parenthesisStarts = array_keys($tokens[$equalsTokenPointer]['nested_parenthesis'] ?? []);
+			$insideParenthesis = max($parenthesisStarts);
```
kukulich: `max()` does not work with empty array: https://3v4l.org/vhUXS

SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
```diff
@@ -54,11 +62,36 @@ public function process(File $phpcsFile, $conditionStartPointer): void
...
+				continue;
+			}
+
+			$functionCall = TokenHelper::findPrevious($phpcsFile, T_STRING, $insideParenthesis, $parenthesisOpener);
+			if ($functionCall !== null) {
+				//this is assignment inside function call
+				continue;
+			}
+
+			//this is assignment inside nested parenthesis
```
kukulich: Remove all inline comment please.

tests/Sniffs/ControlStructures/AssignmentInConditionSniffTest.php
```diff
@@ -17,7 +17,31 @@ public function testCorrectFile(): void
 	public function testIncorrectFile(): void
 	{
 		$resultFile = self::checkFile(__DIR__ . '/data/allAssignmentsInConditions.php');
-		foreach (range(3, 6) as $lineNumber) {
+		foreach (range(3, 7) as $lineNumber) {
+			self::assertSniffError($resultFile, $lineNumber, AssignmentInConditionSniff::CODE_ASSIGNMENT_IN_CONDITION);
+		}
+	}
+
+	public function testNoErrorsWithignoreAssignmentsInsideFunctionCalls(): void
```
kukulich: Typo, use `WithIgnore`

tests/Sniffs/ControlStructures/AssignmentInConditionSniffTest.php
```diff
@@ -17,7 +17,31 @@ public function testCorrectFile(): void
...
+		$report = self::checkFile(
+			__DIR__ . '/data/noAssignmentsInConditionsIgnoreAssignmentsInsideFunctionCalls.php',
+			[
+				'ignoreAssignmentsInsideFunctionCalls' => true,
+			]
+		);
+		self::assertNoSniffErrorInFile($report);
+	}
+
+	public function testErrorsWithignoreAssignmentsInsideFunctionCalls(): void
```
kukulich: Typo, use `WithIgnore`

SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
```diff
@@ -54,11 +62,36 @@ public function process(File $phpcsFile, $conditionStartPointer): void
...
 			return;
 		}
 
+		foreach ($equalsTokenPointers as $equalsTokenPointer) {
+			$parenthesisStarts = array_keys($tokens[$equalsTokenPointer]['nested_parenthesis'] ?? []);
+			$insideParenthesis = max($parenthesisStarts);
+			if ($insideParenthesis === $parenthesisOpener) {
+				$this->error($phpcsFile, $conditionType, $equalsTokenPointer);
+				//this is assignment inside if expression
+				continue;
```
kukulich: The `continue` makes sense only if we report all assignments in one condition. Can you please add test that reports it?

Something like `if ($a = 1 && $b = 1)`

SlevomatCodingStandard/Sniffs/ControlStructures/AssignmentInConditionSniff.php
nightlinus: this is true, but we are inside if so nested_parenthesis will be always at least one element, i can add check for empty array, but it will never be hitted by code coverage

kukulich: So ` ?? []` is useless ?

nightlinus: Yep, it is for phpstan because nested_parenthesis key is optional

kukulich: It can be ignored in https://github.com/slevomat/coding-standard/blob/master/build/PHPStan/phpstan.neon then.

PR #960, opened 2020-04-01
LineLengthSniff: speedup by caching use statements

nightlinus:
Introduce use statement pointers cache for UserStatementHelper.
Speedup for LineLengthSniff on test file from 0.7s to 0.3s


kukulich:
Thanks.

PR #963, opened 2020-04-01
Fix `Classes.ParentCallSpacing` documentation

mathieu-bour:
Changed the probably misstyped `linesCountBeforeFirstControlParentCalle` sniff property name to `linesCountBeforeFirstParentCall`.


kukulich:
Thanks.

PR #987, opened 2020-04-30
FilepathNamespaceExtractor must omit current working dir

rmikalkenas:
Currently TypeNameMatchesFileNameSniff fails if file path is `/www/src/src/Foo/FooTest.php` and defined root namespace is `'src' => 'Slevomat'`. By removing current working dir path from files path, we can assure that filepath namspace extractor will do its job correctly.


c33s:
@rmikalkenas is there a reason why this wasn't merged? i currently have exactly this problem. see #1249

PR #989, opened 2020-05-02
Added tentative support for php8

staabm:
In the process of getting composer/composer ready for php8 , see https://github.com/composer/semver/pull/91

- [ ] requires https://github.com/slevomat/coding-standard/pull/990


kukulich:
Last version can be installed on PHP 8. Tests are not running on PHP 8 because there's a problem with PHPUnit.

So closing.

PR #990, opened 2020-05-02
update to latest shivammathur/setup-php Github Action

staabm:
using the recommended style of depdency declaration promoted by the Github Action author.

also getting a current php8 version with it.


kukulich:
Thanks.

PR #992, opened 2020-05-03
Fix: exclude provided base path in TypeNameMatchesFileNameSniff

rmikalkenas:
TypeNameMatchesFileNameSniff fails if app code is in the same directory as analyzed directory, threfore base path must be omited if provided

### DIFF
--- a/tests/Sniffs/Files/TypeNameMatchesFileNameSniffTest.php
+++ b/tests/Sniffs/Files/TypeNameMatchesFileNameSniffTest.php
@@ -84,4 +84,11 @@ class TypeNameMatchesFileNameSniffTest extends TestCase
 		self::assertNoSniffErrorInFile(self::checkFile(__DIR__ . '/data/rootNamespace2/Xxx/Boo.php', $sniffProperties2));
 	}
 
+	public function testWithProvidedBasePathAndNestedSameDirName(): void
+	{
+		self::assertNoSniffErrorInFile(self::checkFile(__DIR__ . '/data/data/Foo/Bar.php', [
+			'rootNamespaces' => ['data' => 'Data'],
+		], [], ['--basepath=' . __DIR__ . '/data']));
+	}
+
 }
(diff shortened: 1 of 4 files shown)


rmikalkenas:
Not sure why test fails on windows build.. I don't have windows where I could replicate the error, maybe someone from maintainers could help here?

kukulich:
@rmikalkenas 

Filename: P:\Git\slevomat-coding-standard\tests\Sniffs\Files/data/data/Foo/Bar.php
Basepath: P:\Git\slevomat-coding-standard\tests\Sniffs\Files\data

You have to normalize slashes in both paths to `DIRECTORY_SEPARATOR`.

tests/Sniffs/Files/TypeNameMatchesFileNameSniffTest.php
```diff
@@ -84,4 +86,14 @@ public function testRootNamespacesNormalization(): void
 		self::assertNoSniffErrorInFile(self::checkFile(__DIR__ . '/data/rootNamespace2/Xxx/Boo.php', $sniffProperties2));
 	}
 
+	public function testWithProvidedBasePathAndNestedSameDirName(): void
+	{
+		$filePath = str_replace('/', DIRECTORY_SEPARATOR, __DIR__ . '/data/data/Foo/Bar.php');
```
kukulich, line 91: @rmikalkenas This is wrong. You need to normalize it in `TypeNameMatchesFileNameSniff.php`, line 93.

rmikalkenas: are you sure? Because `$phpcsFile->getFilename();` is already normalized by `squizlabs/php_codesniffer`.
I think it should be normalized at test, because filepath Im writing is with forwardslash and windows uses backslack

kukulich: Yes, I'm sure. You cannot rely on PHPCS behaviour. It should be normalized just before the check.

kukulich:
@rmikalkenas Btw please use rebase and squash everything to one commit.

PR #1053, opened 2020-07-06
Useless variable false-positive with trigger_error()

leofeyer:
I have discovered a strange false-positive in the useless variable check in conjunction with `trigger_error()`. So far, the PR only contains a failing unit test to outline the problem. Without the `trigger_error()` call, the test does not fail.

@m-vo /cc


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/c4c391b305fe2453dfbc7c54b25df074048fa848

leofeyer:
Thank you @kukulich. 👍🏻

PR #1106, opened 2020-10-16
Fixed TOC

zf2timo:
In the Table of content was a "Using all sniffs from the standard" Headline. This didn't exists anymore. 
I replaced it with the new "Exclude sniffs you don't want to run" section


kukulich:
Thanks.

Issue #1099, opened 2020-10-13
Adding some of phpcs-calisthenics-rules to this package


olivernybroe:
Since phpcs-calisthenics-rules has been deprecated, I was thinking if this package would be interested in taking over the maintain-work of some of their sniffs?

I think they had a bunch of valuable sniffs which would be awesome to have in the coding-standard.

https://github.com/object-calisthenics/phpcs-calisthenics-rules

kukulich:
I can probably accept some of that sniffs but someone has to prepare PR with tests - one for each sniff.

PR #1108, opened 2020-10-22
Add a sniff to specify maximum number of properties per class

50bhan:
Following issue #1099.

### DIFF
--- a/SlevomatCodingStandard/Helpers/ClassHelper.php
+++ b/SlevomatCodingStandard/Helpers/ClassHelper.php
@@ -3,13 +3,16 @@
 namespace SlevomatCodingStandard\Helpers;
 
 use PHP_CodeSniffer\Files\File;
+use function array_filter;
 use function array_merge;
 use function array_reverse;
+use function count;
 use function sprintf;
 use const T_ANON_CLASS;
 use const T_FINAL;
 use const T_STRING;
 use const T_USE;
+use const T_VARIABLE;
 
 /**
  * @internal
@@ -105,6 +108,30 @@ class ClassHelper
 		}
 
 		return $useStatements;
+	}
+
+	public static function getPropertiesCount(File $file, int $position): int
+	{
+		return count(self::getProperties($file, $position));
+	}
+
+	/**
+	 * @param File $file
+	 * @param int $position
+	 * @return int[]
+	 */
+	public static function getProperties(File $file, int $position): array
+	{
+		$tokens = $file->getTokens();
+		$token = $tokens[$position];
+		$pointer = $token['scope_opener'];
+
+		return array_filter(
+			TokenHelper::findNextAll($file, T_VARIABLE, $pointer + 1, $token['scope_closer']),
+			static function ($variablePointer) use ($file): bool {
+				return PropertyHelper::isProperty($file, $variablePointer);
+			}
+		);
 	}
 
 	/**
--- a/SlevomatCodingStandard/Sniffs/Classes/PropertyPerClassLimitSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/PropertyPerClassLimitSniff.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\Classes;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\ClassHelper;
+use SlevomatCodingStandard\Helpers\SniffSettingsHelper;
+use function sprintf;
+use const T_ANON_CLASS;
+use const T_CLASS;
+use const T_TRAIT;
+
+final class PropertyPerClassLimitSniff implements Sniff
+{
+
+	public const CODE_PROPERTY_PER_CLASS_LIMIT = 'PropertyPerClassLimit';
+
+	/** @var int */
+	public $maxPropertyCount = 10;
+
+	/**
+	 * @return array<int, (int|string)>
+	 */
+	public function register(): array
+	{
+		return [T_CLASS, T_TRAIT, T_ANON_CLASS];
+	}
+
+	/**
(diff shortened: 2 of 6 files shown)


SlevomatCodingStandard/Helpers/ClassHelper.php
```diff
@@ -104,6 +112,39 @@ public static function getTraitUsePointers(File $phpcsFile, int $classPointer):
...
+		$token = $tokens[$position];
+		$pointer = $token['scope_opener'];
+
+		self::$propertyList = [];
+
+		while (($pointer = $file->findNext(T_VARIABLE, $pointer + 1, $token['scope_closer'])) !== false) {
+			self::extractPropertyIfFound($file, $pointer);
+		}
+
+		return self::$propertyList;
```
aadmathijssen, line 138: I think this code be rewritten as follows:

```php
		return array_filter(
			TokenHelper::findNextAll($file, T_VARIABLE, $pointer + 1, $token['scope_closer']),
			static function ($variablePointer) use ($file) {
				return PropertyHelper::isProperty($file, $variablePointer);
			}
		);
```

This avoids the use of a static variable, and reuses the `TokenHelper::findNextAll` method.

SlevomatCodingStandard/Sniffs/Classes/PropertyPerClassLimitSniff.php
```diff
@@ -0,0 +1,53 @@
...
+
+	/** @var int */
+	public $maxCount = 10;
+
+	/**
+	 * @return array<int, (int|string)>
+	 */
+	public function register(): array
+	{
+		return [T_CLASS, T_TRAIT];
```
aadmathijssen, line 26: By adding `T_ANON_CLASS` to this array, this sniff can also be used for anonymous classes. I don't think any extra changes are required (except for adding a few test cases maybe).

SlevomatCodingStandard/Sniffs/Classes/PropertyPerClassLimitSniff.php
```diff
@@ -0,0 +1,53 @@
...
+use const T_CLASS;
+use const T_TRAIT;
+
+final class PropertyPerClassLimitSniff implements Sniff
+{
+
+	public const CODE_PROPERTY_PER_CLASS_LIMIT = 'PropertyPerClassLimit';
+
+	/** @var int */
+	public $maxCount = 10;
```
aadmathijssen, line 19: I suggest to rename this property to `$maxPropertyCount`, because it is more consistent with the naming of other count/length/limit properties in this repository.

aadmathijssen:
Hi @50bhan and @kukulich,

I created a separate PR for the related sniff that limits the number of methods (instead of properties): !1214.

After that I took a look at this PR, and saw some room for minor improvements (see the individual code comments above).

Sorry for the spam. 😀

Aad

50bhan:
@aadmathijssen Thank you for your review and good points! Agree with all of them and made the changes 👍

SlevomatCodingStandard/Helpers/ClassHelper.php
```diff
@@ -107,6 +110,30 @@ public static function getTraitUsePointers(File $phpcsFile, int $classPointer):
...
+		$tokens = $file->getTokens();
+		$token = $tokens[$position];
+		$pointer = $token['scope_opener'];
+
+		return array_filter(
+			TokenHelper::findNextAll($file, T_VARIABLE, $pointer + 1, $token['scope_closer']),
+			static function ($variablePointer) use ($file): bool {
+				return PropertyHelper::isProperty($file, $variablePointer);
+			}
+		);
```
aadmathijssen, line 134: I discovered that this method (the current as well as the previous implementation) does not take the possibility into account that this class itself might define anonymous classes with their own properties. Currently these properties are added to the list of class properties, which I think is incorrect. As an example, in the following class, the number of class properties is counted as 11:

```php
<?php

class FineAmountOfPropertiesWithNestedPropertiesClass
{
	protected $property1;
	protected $property2;
	protected $property3;
	protected $property4;
	protected $property5;
	protected $property6;
	protected $property7;
	protected $property8;
	protected $property9;
	protected $property10;

	protected function anonymousClassMethod()
	{
		return new class {
			protected $anonymousProperty1;
		};
	}
}
```

To resolve this, you can use the `level` of a token: all property tokens should be exactly one level higher than the level of the scope opener token.

I tried to first this into the above implementation above, but the closure started to work against me.

So I ended up with the following:

```php
		$tokens = $file->getTokens();
		$token = $tokens[$position];
		$scopeOpenerPointer = $token['scope_opener'];
		$scopeCloserPointer = $token['scope_closer'];
		$acceptedLevel = $tokens[$scopeOpenerPointer]['level'] + 1;

		$propertyPointers = [];
		foreach (TokenHelper::findNextAll($file, T_VARIABLE, $scopeOpenerPointer + 1, $scopeCloserPointer) as $variablePointer) {
			if (($tokens[$variablePointer]['level'] === $acceptedLevel) && PropertyHelper::isProperty($file, $variablePointer)) {
				$propertyPointers[] = $variablePointer;
			}
		}
		return $propertyPointers;
```

SlevomatCodingStandard/Sniffs/Classes/PropertyPerClassLimitSniff.php
```diff
@@ -0,0 +1,54 @@
...
+			return;
+		}
+
+		$tokenType = $file->getTokens()[$classPointer]['content'];
+		$errorMessage = sprintf(
+			'"%s" has too many properties: %d. Can be up to %d properties.',
+			$tokenType,
+			$propertiesCount,
+			$this->maxPropertyCount
+		);
```
aadmathijssen, line 49: Hi @50bhan, I don't know if this is an actual issue or not, but the original value of `$this->maxPropertyCount` is used in this `sprintf` command, but is converted to an integer using the `%d` specifier. This conversion may or may not be different from that of `SniffSettingsHelper::normalizeInteger`.

In the error printing of sniff properties in most other sniffs, the normalized value is passed to the`sprintf`.

aadmathijssen:
Hi @kukulich, can you please indicate why this PR is closed? Thanks. Aad

PR #1111, opened 2020-10-28
RequireMultiLineTernaryOperatorSniff: added minimal length of 'then & else' expressions

dg:
It is useful to prevent transformation when 'then & else' part is really short, like this

```php
$stats->coveredMethodCount += $lineCount === $coveredLineCount ? 1 : 0;
```

to

```php
$stats->coveredMethodCount += $lineCount === $coveredLineCount
	? 1
	: 0;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,6 +648,7 @@ Ternary operator has to be reformatted to more lines when the line length exceed
 Sniff provides the following settings:
 
 * `lineLengthLimit` (defaults to `0`)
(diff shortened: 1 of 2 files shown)


kukulich:
Implemented with tests in https://github.com/slevomat/coding-standard/commit/a71efd81d8f572e6c55305c689dc4f037dadb6ae

dg:
Friday evenings with open source :)

PR #1114, opened 2020-10-30
RequireCombinedAssignmentOperatorSniff: prevents fatal error

dg:
Prevents fatal error 'Cannot use assign-op operators with string offsets' when changes `$m[0] = $m[0] | 'x'` to `$m[0] |= $m[0]'`


kukulich:
Improved in https://github.com/slevomat/coding-standard/commit/fdb837d016517d2dca85716428231232a7bcb8ca

Error will be reported but not fixed.

I cannot recognize in PHPCS.

dg:
It still reports error: `Use "|=" operator instead of "=" and "|"`

PR #1157, opened 2021-01-29
AbstractPropertyAndConstantSpacing::process(): Next pointer can be null

janbarasek:
Variable `$nextPointer` can be null (and internal logic ignored):

Snímek obrazovky 2021-01-29 v 19 32 50

Thanks.


kukulich:
@janbarasek I don't see any test case for this.

kukulich:
Closed because of no response.

PR #1160, opened 2021-02-04
Wrong resolving of the first type from union type list

bnowak:
I use your standard (which is vey cool) and also `SlevomatCodingStandard.Namespaces.UnusedUses` sniff.

Today I noticed that it's not working properly with the PHP8 union types.

```php
use App\ClassA;
use App\ClassB;

class SomeClass
{
    public function someMethod(ClassA|ClassB $someParam)
    {
    }    
}
```
In the above example `ClassB` is resolved properly as a class type, but `ClassA` is resolved as a constant type which is wrong. They both should be class types.

I've made fix and unit test adjustment to cover this case.

### DIFF
--- a/SlevomatCodingStandard/Helpers/ReferencedNameHelper.php
+++ b/SlevomatCodingStandard/Helpers/ReferencedNameHelper.php
@@ -276,6 +276,10 @@ class ReferencedNameHelper
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/790822f62ab4209d88459be75de780f1734440b5

PR #1169, opened 2021-02-24
Make sure we have a classPointer to fetch a token for

StijnVrolijk:
When `ClassHelper::getClassPointer` returns `null`, the next statement would fail because we would attempt to fetch `$tokens[null]`.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/MethodSpacingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/MethodSpacingSniff.php
@@ -54,6 +54,9 @@ class MethodSpacingSniff implements Sniff
 			: TokenHelper::findNext($phpcsFile, T_SEMICOLON, $methodPointer + 1);
 
 		$classPointer = ClassHelper::getClassPointer($phpcsFile, $methodPointer);
+		if ($classPointer === null) {
+			return;
+		}
 
 		$nextMethodPointer = TokenHelper::findNext($phpcsFile, T_FUNCTION, $methodEndPointer + 1, $tokens[$classPointer]['scope_closer']);
 		if ($nextMethodPointer === null) {


kukulich (changes requested):
I don't see any test case for this situation.

StijnVrolijk:
The code that led me to this issue was the following (trimmed to the most basic example):

errorfile.php
```php
<?php

namespace Foo;

class Bar
{
    public function lorem(): void
    {
        function ipsum(): void
        {
            // This nested function causes an exception
        }
    }
}
```

ruleset.xml
```xml
<?xml version="1.0"?>
<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="vendor/squizlabs/php_codesniffer/phpcs.xsd">
    <rule ref="SlevomatCodingStandard.Classes.MethodSpacing" />
</ruleset>
```

command
```bash
phpcs --standard=ruleset.xml errorfile.php
```

I know nested functions are a big no no but it is valid PHP. I just found this piece of code in an old project I've been attempting to clean up.

Anyway, this causes an undefined index error on SlevomatCodingStandard/Sniffs/Classes/MethodSpacingSniff.php:61.

If you would run MethodSpacingSniff through a static analysis tool like PHPStan or psalm it would tell you about this issue. `ClassHelper::getClassPointer` returns `int|null` but on the next line it is used like it's always `int`. By aborting if the pointer is `null` this error gets prevented.

kukulich:
@StijnVrolijk So just write test please :)

StijnVrolijk:
Ah sorry, I misunderstood. I've looked at the tests and unfortunately my time/skill is lacking so I won't be able to write a test for this. Feel free to close this PR, I just wanted to let you know about this :)

kukulich:
Ok, closing because there are not tests.

PR #1189, opened 2021-03-31
Improve performance of searching references

nightlinus:
When searching for previous attribute
we were searching always up to open tag,
so in case of no attributes it was scanning file
many many times. We add restriction to search up to
 previous scanned pointer so that we scan file only once.
 Performance increased by order of magnitude in specific cases.
 Affected sniffs:
 - FullyQualifiedExceptions
 - UnusedUses
 - FullyQualifiedGlobalConstants
 - FullyQualifiedGlobalFunctions
 - ReferenceThrowableOnly


kukulich:
Thanks.

PR #1190, opened 2021-04-01
Improve performance of findNextAll by batching search and caching result

nightlinus:
Performance increased by order of magnitude in specific cases.
 Affected sniffs:
 - RequireSingleLineCall

This allow us to think less when use findNextAll, as it was root of many perf bottlenecks. Possibility of future perf regression will be decreased. Later we can remove unnessesary caching from more high level calls of findNextAll

### DIFF
--- a/SlevomatCodingStandard/Helpers/TokenHelper.php
+++ b/SlevomatCodingStandard/Helpers/TokenHelper.php
@@ -5,7 +5,12 @@ namespace SlevomatCodingStandard\Helpers;
 use PHP_CodeSniffer\Files\File;
 use function array_key_exists;
 use function array_merge;
+use function array_slice;
 use function count;
+use function implode;
+use function is_int;
+use function is_string;
+use function sort;
 use const T_ARRAY_HINT;
 use const T_BREAK;
 use const T_CALLABLE;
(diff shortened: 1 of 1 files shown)


kukulich:
@nightlinus Do you have any data how the performance is improved?

nightlinus:
For RequireSingleLineCall time drops from 7s to 0.5s for my 5k lines test file, other sniffs within stderr.

SlevomatCodingStandard/Helpers/TokenHelper.php
```diff
@@ -123,6 +127,41 @@ public static function findNext(File $phpcsFile, $types, int $startPointer, ?int
 	 * @return int[]
 	 */
 	public static function findNextAll(File $phpcsFile, $types, int $startPointer, ?int $endPointer = null): array
+	{
+		return self::findNextAllBatch($phpcsFile, $types, $startPointer, $endPointer);
```
kukulich, line 131: I don't like the two new `public` methods. And I don't like the `Positions` class either. I would be happy with everything in `findNextAll()` method.

kukulich (changes requested):

SlevomatCodingStandard/Helpers/TokenHelper.php
nightlinus: Will make two new methods private. Are you sure we want to inline Positions logic, it will be much harder to test and slice logic is not trivial, but i can inline it if you wish.

kukulich: You can create more private methods but I don't like any new `public` classes when not neccessary.

kukulich:
Thanks.

nightlinus:
Do you revert it because of bugs or you find better way to optimize sniffs?

kukulich:
This should be better: https://github.com/slevomat/coding-standard/commit/09249ea680c05b5fd5e7e197fc5d4fe9448f5c2c

PR #1212, opened 2021-04-30
Fix: resolve off-by-one error in function length sniff

aadmathijssen:
In almost all standards, function line counting starts at the line after the opening brace and stops at the line before the closing brace.

So in case the opening brace is on line `l` and the closing brace is on line `k`, we need to count lines `l+1` up to and including `k-1`, which is `(k-1) - (l+1) + 1`, which is the same as `k - l - 1`.


kukulich:
Thanks.

PR #1214, opened 2021-04-30
Feature: add a sniff to limit the number of class, trait and interface methods

aadmathijssen:
This PR adds the `SlevomatCodingStandard.Classes.MethodPerClassLimit` sniff which allows the user to limit the number of allowed methods per class, trait, interface or anonymous class. The number of allowed methods can be configured via the `maxMethodCount` property which defaults to `10`.

This sniff can be used as a replacement for the abandoned [`ObjectCalisthenics.Metrics.MethodPerClassLimit`](https://github.com/object-calisthenics/phpcs-calisthenics-rules/blob/master/src/ObjectCalisthenics/Sniffs/Metrics/MethodPerClassLimitSniff.php) sniff. However, that sniff does not support anonymous classes, or functions defined within class methods. Both are supported by the sniff in this PR.

This PR is related to issue #1099.


aadmathijssen:
Hi @kukulich, can you please indicate why this PR is closed? Thanks. Aad

PR #1217, opened 2021-05-03
Fix: resolve false positive in FunctionHelper::isMethod for functions defined within a class method

aadmathijssen:
The helper function `\SlevomatCodingStandard\Helpers\FunctionHelper::isMethod` incorrectly returns `true` for functions that are defined within a class method.

Take the following example:

```php
class Foo
{
    public function classMethod()
    {
        function nonClassMethodFunction()
        {
        }

        return nonClassMethodFunction();
    }
}
```

Here, `FunctionHelper::isMethod` function returns `true` for `classMethod` as well as `nonClassMethodFunction`, while for the latter it should return `false`.

This PR fixes this issue.

### DIFF
--- a/SlevomatCodingStandard/Helpers/FunctionHelper.php
+++ b/SlevomatCodingStandard/Helpers/FunctionHelper.php
@@ -9,6 +9,7 @@ use SlevomatCodingStandard\Helpers\Annotation\ReturnAnnotation;
 use function array_filter;
 use function array_map;
 use function array_merge;
+use function array_pop;
 use function array_reverse;
 use function count;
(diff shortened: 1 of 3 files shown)


kukulich:
Oh, really ugly situation but thanks :)

aadmathijssen:
You're welcome. :-) I would never use this in practice myself, but it seems to be valid PHP.
BTW I found this out while testing #1214, and I was looking for a use of `T_FUNCTION` within a class that was a not a method. It looks like this is the only possible scenario (anonymous functions and closures have different tokens).

PR #1222, opened 2021-05-08
ReferenceUsedNamesOnly: fix requiring use of global names

schlndh:
When allowFullyQualifiedGlobalX is set to false it should require global X to be used. However, this was not the case when used together with non-empty namespacesRequiredToUse.

My use-case for this is that I want to require using my project's namespace and the global namespace.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Namespaces/ReferenceUsedNamesOnlySniff.php
+++ b/SlevomatCodingStandard/Sniffs/Namespaces/ReferenceUsedNamesOnlySniff.php
@@ -242,7 +242,7 @@ class ReferenceUsedNamesOnlySniff implements Sniff
 			}
 
 			if ($isFullyQualified || $isGlobalFunctionFallback || $isGlobalConstantFallback) {
-				if ($isFullyQualified && !$this->isRequiredToBeUsed($name)) {
+				if ($isFullyQualified && !$this->isRequiredToBeUsed($reference)) {
 					continue;
 				}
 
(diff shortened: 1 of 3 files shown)


kukulich:
I think there are other sniffs that do this.

schlndh:
@kukulich I skimmed through the list of sniffs in README and haven't found any. In either case, the behavior of `ReferenceUsedNamesOnlySniff` is a bit unfortunate

> namespacesRequiredToUse: if not set, all namespaces are required to be used. When set, only mentioned namespaces are required to be used. Useful in tandem with UseOnlyWhitelistedNamespaces sniff.
> allowFullyQualifiedGlobalClasses: allows using fully qualified classes from global space (i.e. \DateTimeImmutable).

Let's go through the possibilities:

- `namespacesRequiredToUse = []`, `allowFullyQualifiedGlobalClasses = false`: Everything is required to be used. This is the default and it works as expected.
- `namespacesRequiredToUse = []`, `allowFullyQualifiedGlobalClasses = true`: Everything but classes in the global namespace is required to be used.
- `namespacesRequiredToUse = ['Foo']`, `allowFullyQualifiedGlobalClasses = false`: Only stuff from `Foo` is required to be used. I would expect that global classes must be used as well, however it seems that in this case the option is ignored.
- `namespacesRequiredToUse = ['Foo']`, `allowFullyQualifiedGlobalClasses = true`: Only stuff from `Foo` is required to be used.

Currently, the behavior of the last 2 possibilites is the same. It seems to me that my suggestion enhances the capabilities of the sniff. It was not possible to use a non-empty namespacesRequiredToUse AND require global classes to be used at the same time. Unfortunately, since false is the default for `allowFullyQualifiedGlobal*` projects that use the sniff with a non-empty namespacesRequiredToUse and did not want to require using stuff from global namespace would have to change the settings to true.

kukulich:
You're probably looking for:

- `UseOnlyWhitelistedNamespacesSniff`
- `FullyQualifiedGlobalConstantsSniff`
- `FullyQualifiedGlobalFunctions`

Yes, it's little complicated. Some options were added earlier, some later - it's not intuitive sometimes. I'll be refactored in next major version.

schlndh:
@kukulich I don't think those sniffs address my problem. I'm trying to setup a rule like this:

- Everything from namespace `Foo` must be used.
- Everything from the global namespace must be used. `FullyQualifiedGlobal*Sniff` is not useful here because they do the "opposite".
- Everything else may or may not be used depending on the situation. This rules out `UseOnlyWhitelistedNamespacesSniff`.

PR #1266, opened 2021-08-25
Add DisallowNonCapturingCatchSniff

olsavmic:
I think there is a good reason to forbid PHP 8 non-capturing catch as the new language sugar may lead developers into discarding exceptions instead of working with them properly (eg. logging).

The sniff does not contain automatic fixer as the fix requires manual work - if it was fixed automatically, it'd be no different to non-capturing catch block.


janedbal:
@kukulich Is there any chance this gets released any time soon? What is your release plan?

PR #1271, opened 2021-08-30
ClassStructureSniff: add missing magic methods

jrfnl:
PHP 7.4 introduced `__serialize()` and `__unserialize()` as magic methods.


kukulich:
Thanks!

PR #1273, opened 2021-09-02
Support @covers annotation

ptahjke:
annotation `@covers` can look like someClass::someMethod  -- eh it's non recommended but it can be
and it will be great to identify import from `@covers` annotation :)


kukulich:
@ptahjke Can you please add test?

Issue #1265, opened 2021-08-25
False positive SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName


dereuromark:
We would like to enable this sniff, but the side effects right now killing our code a bit too high
It seems to also remove some inside docblocks, which should not happen (they are FQCN inside for a reason), but it also seems to kill code inside strings.

This is an example patch of the changes it did to our code:

```diff
@@ -23,6 +23,7 @@
 use Spryker\Shared\Config\Config;
 use Spryker\Shared\ErrorHandler\ErrorHandlerEnvironment;
 use Spryker\Shared\PropelOrm\PropelOrmConstants;
+use Spryker\Zed\Kernel\Persistence\EntityManager\TransferToEntityMapper;
 
 class ObjectBuilder extends PropelObjectBuilder
 {
@@ -108,8 +109,7 @@
         // When this is true we will not check for value equality as we need to be able to set a value for this field
         // to it's initial value and have the column marked as modified. This is relevant for update cases when
         // we create an instance of an entity manually.
-        // @see \Spryker\Zed\Kernel\Persistence\EntityManager\TransferToEntityMapper::mapEntity()
-        \$hasDefaultValue = $hasDefaultValue;
+TransferToEntityMapper        \$hasDefaultValue = $hasDefaultValue;
 
         if (\$hasDefaultValue || \$this->$clo !== \$v) {
             \$this->$clo = \$v;
@@ -151,8 +151,7 @@
         // When this is true we will not check for value equality as we need to be able to set a value for this field
         // to it's initial value and have the column marked as modified. This is relevant for update cases when
         // we create an instance of an entity manually.
-        // @see \Spryker\Zed\Kernel\Persistence\EntityManager\TransferToEntityMapper::mapEntity()
-        \$hasDefaultValue = $hasDefaultValue;
+TransferToEntityMapper        \$hasDefaultValue = $hasDefaultValue;
 
         if (\$hasDefaultValue || \$this->$clo !== \$v) {
             \$this->$clo = \$v;

@@ -22,6 +22,7 @@
 use Spryker\Shared\Config\Config;
 use Spryker\Shared\ErrorHandler\ErrorHandlerEnvironment;
 use Spryker\Shared\PropelOrm\PropelOrmConstants;
+use Spryker\Zed\Kernel\Persistence\EntityManager\TransferToEntityMapper;
 
[...]

PR #1281, opened 2021-10-02
Appended logic for find names of classes inside double quotes

AIlkiv:
Simple case about this pull request

```
use Config;

...

$query = "SELECT * FROM config WHERE type = {$db->quote(Config::COMMON)}";
```

UnusedUsesSniff will say Config is not used.

This pull request fixing it.


AIlkiv:
@kukulich
Fixed.
Add test.
Manual tested phpcs on repository from issue #1265

kukulich:
@AIlkiv Please squash commits to one and it should be ready to merge.

PR #1298, opened 2021-11-04
Makes RequireTrailingCommaInCallSniff check if the closing paren is on the same line

donatj:
Makes RequireTrailingCommaInCallSniff check if the closing paren is on the same line as the item before, and skip it if it is.

It's already doing so if there's no whitespace, but if there's whitespace on the line it adds a comma. I don't imagine this is a desired behaviour.

---

Right now RequireTrailingCommaInCallSniff will leave this code alone:

```php
doSomething($a,
	$b);

doSomething(doSomething(
	$foo,
));
```

but turn these

```php
doSomething( $a,
	$b );

doSomething( doSomething(
	$foo,
) );
```

into

```php
doSomething( $a,
	$b, );

doSomething( doSomething(
	$foo,
), );
```

I've basically fixed it so it leaves these instances alone.


kukulich:
Thanks!

Issue #1304, opened 2021-11-20
DeclareStrictTypesSniff: written change fails subsequent sniff


arderyp:
if I sniff/fix a file without `strict_types` declared, it adds the following to the top of the file:

```
declare(strict_types = 1);
```

When I sniff/fix the same file right after, it warns me about incorrect `strict_types` format, and then implements the same change:

```
[SNIFF]
-----------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------
 3 | ERROR | [x] Expected strict_types = 1, found strict_types=1.
-----------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------------

Time: 25ms; Memory: 6MB


[FIX]

PHPCBF RESULT SUMMARY
--------------------------------------------------------------------------------------------------------------
FILE                                                                                          FIXED  REMAINING
--------------------------------------------------------------------------------------------------------------
<MY-FILE.php>     1      0
--------------------------------------------------------------------------------------------------------------
A TOTAL OF 1 ERROR WERE FIXED IN 1 FILE
--------------------------------------------------------------------------------------------------------------
```

The issue is that its checking for `strict_types = 1` but writing `strict_types=1`.

Incoming PR to fix...

arderyp:
https://github.com/slevomat/coding-standard/pull/1305

kukulich:
You have to configure the `DeclareStrictTypes` sniff.

arderyp:
Im sory @kukulich, I don't follow.  Can you point me to some documentation or explain further?

kukulich:
@arderyp https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintsdeclarestricttypes-

arderyp:
sorry, yeah, I see it now, thanks!

PR #1305, opened 2021-11-20
fixes 1304: write 'strict_types=1' instead of 'strict_types = 1'

arderyp:
fixes 1304


kukulich:
https://github.com/slevomat/coding-standard/issues/1304#issuecomment-974653094

PR #1306, opened 2021-11-22
Add `SlevomatCodingStandard.Classes.RequireAbstractOrFinal`

roslov:


simPod (commented):
Many people use doctrine orm that use proxies - cannot declare those classes final so they can be proxified. Is there any way we could provide a configuration for that use case?

roslov:
> Many people use doctrine orm that use proxies - cannot declare those classes final so they can be proxified. Is there any way we could provide a configuration for that use case?

I also use Doctrine. For the proxied classes (like entities) we can add an exclusion, for example:

```xml
<rule ref="SlevomatCodingStandard.Classes.RequireAbstractOrFinal">
    <exclude-pattern>src/Entity/*</exclude-pattern>
</rule>
```

simPod:
exclude dir works I guess, thanks for reply 👌🏾

roslov:
@kukulich, any updates on this?

kukulich:
Thanks.

PR #1314, opened 2021-12-18
Fix bug in PropertyTypeHintSniff

ondrejmirtes:
If I have `<config name="php_version" value="70100"/>` but I set `<property name="enableUnionTypeHint" value="true"/>` manually, it doesn't work :)


kukulich:
Thanks!

Issue #1358, opened 2022-04-21
False positives in UnusedVariable sniff with increment/decrement operators


aadmathijssen:
Consider the following PHP code:

```php
<?php
function inc(int $a): int
{
    return ++$a;
}

function stop(int $b): void
{
    exit(++$b);
}

function show(int $c): void
{
    print ++$c;
}

function commas(int $d): void
{
    echo $d, ++$d;
}

function one(int $e): bool
{
    return empty(++$e);
}

function evil(int $f): void
{
    eval(++$f . ' !== 1 ?: exit("zero");');
}
```

And the following `phpcs.xml`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ruleset>
    <rule ref="SlevomatCodingStandard.Variables.UnusedVariable"/>
</ruleset>
```

Then phpcs shows the following errors:

```
FILE: /path/to/test.php
-----------------------------------------------------------------------
FOUND 6 ERRORS AFFECTING 6 LINES
-----------------------------------------------------------------------
  4 | ERROR | Unused variable $a.
  9 | ERROR | Unused variable $b.
 14 | ERROR | Unused variable $c.
 19 | ERROR | Unused variable $d.
 24 | ERROR | Unused variable $e.
 29 | ERROR | Unused variable $f.
-----------------------------------------------------------------------```
```

Instead of using the pre-increment expression `++$x`, the other increment/decrement expressions `$x++`, `--$x` and `$x--`  also trigger these errors.

This can all be traced back to the `isPartOfStatementAndWithIncrementOrDecrementOperator` method of the `UnusedVariableSniff` class:

https://github.com/slevomat/coding-standard/blob/7.1.0/SlevomatCodingStandard/Sniffs/Variables/UnusedVariableSniff.php#L650

The highlighted line contains all tokens that are allowed to directly precede the increment/decrement expression. The following tokens are missing: `T_RETURN`, `T_EXIT`, `T_PRINT`, `T_COMMA`, `T_EMPTY`, and `T_EVAL`. Adding these should resolve the reported errors.

kukulich:
Fixed in #1359

PR #1359, opened 2022-04-21
Fix: resolve false positives in UnusedVariable sniff with increment/decrement operators

aadmathijssen:
The `SlevomatCodingStandard.Variables.UnusedVariable` sniff returns false positives when a increment/decrement expression is prefixed with one the following tokens: `T_RETURN`, `T_EXIT`, `T_PRINT`, `T_COMMA`, `T_EMPTY`, and `T_EVAL`.

After merging this PR, the following code should not report any violations against the `SlevomatCodingStandard.Variables.UnusedVariable` sniff anymore:

```php
<?php
function inc(int $a): int
{
    return ++$a;
}

function stop(int $b): void
{
    exit(++$b);
}

function show(int $c): void
{
    print ++$c;
}

function commas(int $d): void
{
    echo $d, ++$d;
}

function one(int $e): bool
{
    return empty(++$e);
}

function evil(int $f): void
{
    eval(++$f . ' !== 1 ?: exit("zero");');
}
```

This PR fixes issue #1358.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Variables/UnusedVariableSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Variables/UnusedVariableSniff.php
@@ -32,7 +32,10 @@ use const T_DOUBLE_COLON;
 use const T_DOUBLE_QUOTED_STRING;
 use const T_ECHO;
 use const T_ELSEIF;
+use const T_EMPTY;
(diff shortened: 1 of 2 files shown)


kukulich:
Thanks.

PR #1360, opened 2022-04-22
Fix: allow increment/decrement operator after yield in UnusedVariable sniff

aadmathijssen:
The `SlevomatCodingStandard.Variables.UnusedVariable` sniff returns a false positive when using an increment/decrement operator on a variable directly after a `yield`.

After merging this PR, the following code should not report any violation against the `SlevomatCodingStandard.Variables.UnusedVariable` sniff anymore:

```php
<?php
function yieldinc($i)
{
    yield ++$i;
}
```


kukulich:
Thanks.

PR #1362, opened 2022-04-28
Add missing header in docs for SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses

herndlm:
noticed by luck while I was searching for the docs of that sniff


kukulich:
Thanks!

PR #1366, opened 2022-05-22
DisallowCommentAfterCodeSniff: don't break multi-line string during auto-fix

schlndh:
There is a bug where
```php
doSomething("
    aaa
"); // comment
```
would be fixed as 
```php
doSomething("
    aaa
// comment
");
```
This is an attempt to fix this scenario.


kukulich:
Looks good, thanks!

PR #1373, opened 2022-05-28
TokenHelper: fix findFirstNonWhitespaceOnLine 1st line

schlndh:
I'm getting issues like this:

```
An error occurred during processing; checking has been aborted. The error
message was: Undefined array key -1 in
.../SlevomatCodingStandard/Helpers/TokenHelper.php
on line 312 (Internal.Exception)
```

The code that triggered this for me had something like this on the first line:

```php
<?php /** @var Abc $abc */ ?>
```

So I also added a test for DisallowCommentAfterCodeSniff.


kukulich:
Thanks.

PR #1410, opened 2022-07-17
SlevomatCodingStandard.Functions.RequireSingleLineCallSniff: skip calls with multi-line double-quoted string

schlndh:
I have code like this:

```php
$table = 'aaa';
$this->query("
    INSERT INTO {$table} (id)
    VALUES (1), (2), (3)
");
```

And I got this:

> Call of method query() should be placed on a single line.

In this particular case, it doesn't really matter, but in general the sniff can't be sure if the whitespace is significant or not, so it should be ignored.


kukulich:
Thanks.

PR #1412, opened 2022-07-20
Reflect current major version in README and composer.json

derrabus:
The master branch contains version 8.x, but the documentation and the branch alias in composer.json apparently haven't been updated yet.


kukulich:
Thanks!

PR #1416, opened 2022-07-22
RequireNowdocSniff: accepts escaped sequences

dg:
Heredoc sequences like `\n` or `\x77` cannot be converted do nowdoc.


kukulich (changes requested):
The build is broken.

dg:
Fixed

kukulich:
Thanks.

PR #1429, opened 2022-09-05
Add "dev" Composer keyword

GaryJones:
As per https://getcomposer.org/doc/04-schema.md#keywords by including "dev" as a keyword in the `composer.json` file, Composer 2.4.0-RC1 and later will prompt users if the package is installed with `composer require` instead of `composer require --dev`. See https://github.com/composer/composer/pull/10960 for more info.

The "phpcs" is added to match several other coding standards packages.


kukulich:
Thanks, I made the some change as suggested in Doctrine CS.

PR #1446, opened 2022-10-11
Add sniff that disallows using $this in a static closure

annervisser:


ondrejmirtes:
You can have this (and many other bugs) reported by PHPStan: https://phpstan.org/r/6d383041-28d9-4d24-b9e0-7e097fb2b061

I don't think we need PHP_CodeSniffer sniff for this scenario.

annervisser:
We're currently running Psalm which doesn't report this issue :disappointed: , hence this Sniff. I agree this is probably better suited to static analysis tools. If this isn't merged here we'll use it from a custom rule for the time being :slightly_smiling_face:

kukulich:
However thank your for your effort.

PR #1447, opened 2022-10-12
DisallowAttributeJoiningSniff

michnovka:


michnovka:
@kukulich I have finished with the basic concepts, since this is my first contribution, I have some questions:

1. check my `tests/Sniffs/Attributes/data/*.fixed.php`. The indentation is totally messed up. This is because I only care about line spacing, I dont care how my fixes treat indentation. My assumption is that other phpcs rules (and fixes) will then format the indentation according to specified standard. **Is that so?**
2. Check `AttributeHelper::$attributableTokens`. The point here was to make a complete list of php symbols (tokens) where attributes can be applied. The list may be incomplete, do you see anything missing?
3. Right now the `AttributeSpacingSniff` is targetting `T_ATTRIBUTE`. I think it would be also possible to target specifically `AttributeHelper::$attributableTokens` and use `findPrevious` to get the list of all applicable attributes. In such case I would likely add some methods to `AttributeHelper` like `getAllAppliedAttributes()` which could be useful for other Attribute sniffs in the future.
4. Can you please explain the difference between `TokenHelper::findXXX` and `File::findXXX`? The wrappers seem to only convert `false` into `null`. I dont see any purpose. Should I avoid the `File::findXXX` altogether for convention?

Thanks for your feedback!

kukulich:
Please send one PR for each sniff.


>     1. check my `tests/Sniffs/Attributes/data/*.fixed.php`. The indentation is totally messed up. This is because I only care about line spacing, I dont care how my fixes treat indentation. My assumption is that other phpcs rules (and fixes) will then format the indentation according to specified standard. **Is that so?**

If I can, I try to solve (or just improve) indentation as well. Please check `IndentationHelper`.

>     2. Check `AttributeHelper::$attributableTokens`. The point here was to make a complete list of php symbols (tokens) where attributes can be applied. The list may be incomplete, do you see anything missing?

I think `T_TRAIT` is missing.

>     3. Right now the `AttributeSpacingSniff` is targetting `T_ATTRIBUTE`. I think it would be also possible to target specifically `AttributeHelper::$attributableTokens` and use `findPrevious` to get the list of all applicable attributes. In such case I would likely add some methods to `AttributeHelper` like `getAllAppliedAttributes()` which could be useful for other Attribute sniffs in the future.

I think `T_ATTRIBUTE` is better because it will trigger less sniff checks than `AttributeHelper::$attributableTokens`, especially `T_VARIABLE` is bad for this.

>     4. Can you please explain the difference between `TokenHelper::findXXX` and `File::findXXX`? The wrappers seem to only convert `false` into `null`. I dont see any purpose. Should I avoid the `File::findXXX` altogether for convention?

Yes, please try to avoid `File::findXXX` if possible.

michnovka:
@kukulich if I push more PRs they will depend on each other because of the AttributeHelper. Since this is basically solving the same github issue, can we have both sniffs in one PR?

kukulich:
@michnovka No, please one sniff per PR.

michnovka:
@kukulich ok, so how do I approach when the 2 PR have some common code (`AttributeHelper`)? Do I include the identical code for helper functions into both PRs? Or do I first submit PR for helper, then when merged work on the sniffs?

michnovka:
@kukulich ok, so this is now just the `DisallowAttributeJoiningSniff` with `AttributeHelper`. This one I consider finished.

I have moved the work on `AttributeSpacing` to https://github.com/slevomat/coding-standard/pull/1449 and there I want to finish with indentation still.

However this sniff is imo ready for merge

SlevomatCodingStandard/Helpers/AttributeHelper.php
```diff
@@ -0,0 +1,85 @@
...
+		T_INTERFACE => T_INTERFACE,
+		T_ENUM => T_ENUM,
+		T_ENUM_CASE => T_ENUM_CASE,
+		T_FUNCTION => T_FUNCTION,
+		T_FN => T_FN,
+		T_CONST => T_CONST,
+	];
+
+	/**
+	 * @return array<int>
```
kukulich, line 40: Please use `list<int>`

SlevomatCodingStandard/Helpers/AttributeHelper.php
```diff
@@ -0,0 +1,85 @@
...
+
+		$attributeNamePointer = TokenHelper::findNext($phpcsFile, T_STRING, $startPointer + 1, $attributeEndPointer);
+		$attributes[] = $attributeNamePointer;
+
+		while ($commaPointer !== null) {
+			$parenthesisOpenPointer = TokenHelper::findNext($phpcsFile, T_OPEN_PARENTHESIS, $startPointer + 1, $attributeEndPointer);
+			$commaPointer = TokenHelper::findNext($phpcsFile, T_COMMA, $startPointer + 1, $attributeEndPointer);
+
+			if ($commaPointer === null) {
+				// no more attributes
```
kukulich, line 67: Is it possible to use early exit here?

SlevomatCodingStandard/Sniffs/Attributes/DisallowAttributeJoiningSniff.php
```diff
@@ -0,0 +1,71 @@
...
+	{
+		$attributes = AttributeHelper::getAttributesInsideAttributeTags($phpcsFile, $attributeOpenPointer);
+		$attributeCount = count($attributes);
+
+		if ($attributeCount === 1) {
+			return;
+		}
+
+		$fix = $phpcsFile->addFixableError(
+			sprintf('%d attributes are joined', $attributeCount),
```
kukulich, line 42: Please add dot.

SlevomatCodingStandard/Helpers/AttributeHelper.php
```diff
@@ -0,0 +1,85 @@
...
+	/** @var array<int|string,int|string> $attributableTokens */
+	public static $attributableTokens = [
+		T_CLASS => T_CLASS,
+		T_VARIABLE => T_VARIABLE,
+		T_INTERFACE => T_INTERFACE,
+		T_ENUM => T_ENUM,
+		T_ENUM_CASE => T_ENUM_CASE,
+		T_FUNCTION => T_FUNCTION,
+		T_FN => T_FN,
+		T_CONST => T_CONST,
```
kukulich, line 36: `T_TRAIT` is missing. And `T_CLOSURE` as well.

SlevomatCodingStandard/Helpers/AttributeHelper.php
```diff
@@ -0,0 +1,85 @@
...
+	 */
+	public static function getAttributesInsideAttributeTags(File $phpcsFile, int $attributeOpenPointer): array
+	{
+		$attributes = [];
+
+		$tokens = $phpcsFile->getTokens();
+
+		if ($tokens[$attributeOpenPointer]['code'] !== T_ATTRIBUTE) {
+			throw new InvalidArgumentException(
+				sprintf('Token %d must be attribute, %s given', $attributeOpenPointer, $tokens[$attributeOpenPointer]['type'])
```
kukulich, line 50: Please add dot.

tests/Helpers/data/classWithJoinedAttributes.php
```diff
@@ -0,0 +1,7 @@
+<?php // lint >= 8.0
```
kukulich, line 1: Please name the file similar as files for other sniffs.

tests/Sniffs/Attributes/data/disallowedAttributeJoiningErrors.fixed.php
```diff
@@ -0,0 +1,20 @@
...
+}
+
+$object = new #[Attribute1] #[Attribute2('var')]
+    #[Attribute3(option: PDO::class, option2: true, option3: 'False')] class () { /* … */ };
+
+#[Attribute1] #[Attribute2('var')]
+#[Attribute3(option: PDO::class, option2: true, option3: 'False')]
+function testFunc($test)
+{
+}
```
kukulich, line 20: Missing end of line. The bug is in more files.

tests/Sniffs/Attributes/data/disallowedAttributeJoiningNoErrors.php
```diff
@@ -0,0 +1,20 @@
...
+}
+
+$object = new #[Attribute1] #[Attribute2('var')]
+    #[Attribute3(option: PDO::class, option2: true, option3: 'False')] class () { /* … */ };
+
+#[Attribute1] #[Attribute2('var')]
+#[Attribute3(option: PDO::class, option2: true, option3: 'False')]
+function testFunc($test)
+{
+}
```
kukulich, line 20: Please test all other missing types, eg. closure, constant or property.

kukulich (changes requested):
Please also add the sniff to README.

tests/Helpers/data/classWithJoinedAttributes.php
kukulich: Ou, you're right.

tests/Sniffs/Attributes/data/disallowedAttributeJoiningNoErrors.php
kukulich: @michnovka There is "eg." in my comment. All `attributableTokens` should be checked.

You should also check that only parameters and properties are ok for `T_VARIABLE`.

kukulich:
@michnovka https://github.com/slevomat/coding-standard/pull/1447#discussion_r995887201

tests/Sniffs/Attributes/data/disallowedAttributeJoiningNoErrors.php
michnovka: @kukulich  hmm, since I added enums into the data sample, lint for 8.0 is failing (needs 8.1). How do we approach this? Do I create special test case for enums with lint target 8.1 and keep the rest under 8.0, or do I just bump the tests to 8.1 for the whole file?

kukulich: > do I just bump the tests to 8.1 for the whole file?

Just bump it for the whole file.

michnovka: Ok I have added an example of invalid attribute and its treated as a comment. And not fixed.

michnovka:
@kukulich for invalid usage of attributes the fix is now not applied. I think thats all, final review?

kukulich (approved):

PR #1467, opened 2022-11-10
Fix `DisallowAttributesJoiningSniff` for Attribute with trailing comma

javer:
Fixes the following exception in `DisallowAttributesJoiningSniff` when the attribute ends with trailing comma (see test), because it's allowed in PHP:
```php
TypeError: SlevomatCodingStandard\Helpers\TokenHelper::getContent(): Argument #2 ($startPointer) must be of type int, null given, called in SlevomatCodingStandard/Helpers/AttributeHelper.php on line 71

SlevomatCodingStandard/Helpers/TokenHelper.php:426
SlevomatCodingStandard/Helpers/AttributeHelper.php:71
SlevomatCodingStandard/Sniffs/Attributes/DisallowAttributesJoiningSniff.php:35
vendor/squizlabs/php_codesniffer/src/Files/File.php:498
vendor/squizlabs/php_codesniffer/src/Files/LocalFile.php:92
SlevomatCodingStandard/Sniffs/TestCase.php:68
tests/Sniffs/Attributes/DisallowAttributesJoiningSniffTest.php:12
```

### DIFF
--- a/SlevomatCodingStandard/Helpers/AttributeHelper.php
+++ b/SlevomatCodingStandard/Helpers/AttributeHelper.php
@@ -58,6 +58,11 @@ class AttributeHelper
 
 		do {
 			$attributeNameStartPointer = TokenHelper::findNextEffective($phpcsFile, $actualPointer + 1, $attributeCloserPointer);
+
+			if ($attributeNameStartPointer === null) {
+				break;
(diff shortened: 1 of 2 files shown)


kukulich:
Thanks.

PR #1473, opened 2022-11-30
UnusedUse bug when import is used in a docBlock

### LINKED ISSUE #1472: ReferenceViaFullyQualifiedName and UnusedUse sniffs are clashing when use statements are referencing entities in the docBlocks
I have an issue when a class is referenced in a docBlock, but not explicitly used in the code.

This code will produce the following error

 ```php
use WP_Error;
use WP_REST_Request ;
...
	/**
	 * Rest callback
	 * 	 
	 * @param WP_REST_Request $request Data got from endpoint url.
	 *
	 * @return WP_Error|\WP_HTTP_Response|\WP_REST_Response If response generated an error, WP_Error, if response
	 *                                is already an instance, WP_HTTP_Response, otherwise
	 *                                returns a new WP_REST_Response instance.
	 */
	public function routeCallback(WP_REST_Request $request) { ...
```

```bash
Class \WP_HTTP_Response should not be referenced via a fully qualified name, but via a use
     |       |     statement.
     |       |     (SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName)
Class \WP_REST_Response should not be referenced via a fully qualified name, but via a use
     |       |     statement.
     |       |     (SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName)
```

But when I add the use statement at the top of the file I get

```bash
Type WP_HTTP_Response is not used in this file.
    |       |     (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
```

Because I'm not actually using it in my code (only in the docblock so that phpstan can do its magic). I can add `// phpcs:ignore` but the issue is still that this is being falsely flagged.

dingo-d:
Closes #1472 

I've made a test for the bug described in the issue. So far no fix has been done on this, this is just to showcase the issue I am experiencing with the particular sniff.

### DIFF
--- a/tests/Sniffs/Namespaces/UnusedUsesSniffTest.php
+++ b/tests/Sniffs/Namespaces/UnusedUsesSniffTest.php
@@ -281,4 +281,14 @@ class UnusedUsesSniffTest extends TestCase
 		self::assertNoSniffErrorInFile($report);
 	}
 
+	public function testUsesInDocBlock(): void
+	{
+		$report = self::checkFile(__DIR__ . '/data/collidingUnusedUse.php', [
+			'searchAnnotations' => true,
+			'allowFullyQualifiedGlobalFunctions' => true,
+			'allowFullyQualifiedGlobalConstants' => true,
+		], [UnusedUsesSniff::CODE_UNUSED_USE]);
+
+		self::assertSniffError($report, 6, UnusedUsesSniff::CODE_UNUSED_USE, 'Type Baz is not used in this file.');
+	}
 }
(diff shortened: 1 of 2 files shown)


kukulich:
@dingo-d The test is strange. You're expecting bug but there's no bug. The used class is detected as expected.

dingo-d:
@kukulich I've never used your standards to write a test (it's different from the default phpcs ones), I'm not sure what I should test for 😬

I can reproduce phpcs issue, but not sure how to reproduce the test issue 🤷🏼‍♂️

PR #1481, opened 2022-12-22
Add new optimized global functions to FullyQualifiedGlobalFunctionsSniff

fancyweb:
See https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/fb41eb66a32dc186f8068c134055701ac272b217/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L350 and https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/6277.

I also reordered the list.


kukulich:
Thanks.

PR #1482, opened 2022-12-23
Add IntersectionTypeHintSniff

fancyweb:
Add `IntersectionTypeHintSniff` to control the whitespaces of intersection types.

I copied and cleaned all the logic from `UnionTypeHintFormatSniff` and its tests.


kukulich:
Thanks you for your effort however I don't think this sniff makes sense. Since PHP 8.2 we have https://wiki.php.net/rfc/dnf_types so the `UnionTypeHintFormatSniff` should be deprecated and we will need one sniff that has to check both union and intersection types.

kukulich:
> DNF is currently not taken into account completely, isn't it?

Yes, no support here currently.

> I agree a new sniff should replace both UnionTypeHintFormatSniff and IntersectionTypeHintSniff for PHP 8.2. But that doesn't make IntersectionTypeHintSniff invalid for PHP 8.1 IMO.

Yes, but it's better to maintain one sniff that supports all DNF types that have three sniffs.

PR #1484, opened 2022-12-29
Pass error message data to PHPCS in fully qualified reference sniffs

benvoynick:
AbstractFullyQualifiedGlobalReference->process() makes its own call to sprintf() and then passes on that finalized error message to ->addFixableError. This means that customizations of the message for these sniffs cannot use the name of the function/constant via %s.

Changing the call to let PHPCS do the formatting sends this data downstream for possible customization.


kukulich:
I must say - I don't see no extra advantage to do this... However I'm willing to merge the PR if the PR make the change in all sniffs.

PR #1497, opened 2023-01-16
Support "//" comments in AttributeAndTargetSpacingSniff

mmoll:


kukulich:
`//` is not doc comment so it's not valid for this sniff.

mmoll:
> `//` is not doc comment so it's not valid for this sniff.

Alright, now I found out, that the problem is indeed in a different sniff, but at the moment I'm a bit lost how to solve this, apart from the failing test.

Issue #1501, opened 2023-01-23
[SNIFF REQUEST] sort uses by length


andrewalkermo:
There already has a sniff for order imports alphabetically. But i would like another one to sort imports by their length.

Ex.:

```php
use LogableTrait;
use LogAware;
use LogFactory;
use LoggerInterface;
use LogLevel;
use LogStandard;
```
Becomes:

```php
use LogAware;
use LogLevel;
use LogFactory;
use LogStandard;
use LogableTrait;
use LoggerInterface;
```

kukulich:
Thank you for your effort but we don't think this sniff is useful.

PR #1503, opened 2023-01-23
add sniff to order uses by length

andrewalkermo:
Closes: slevomat/coding-standard#1501


kukulich:
Thank you for your effort but we don't think this sniff is useful.

PR #1513, opened 2023-02-02
RequireMultiLineMethodSignatureSniff: allow setting minimal param count

janedbal:
~This is a draft.~ Would you accept such feature if I add proper tests? Default value might be adjusted / removed.

We want to force multi-line signatures for 2+ parameters.


kukulich:
@janedbal Yes, I'm ok with the parametr. However I think both parameters `minLineLength` and `minParameterCount` should not be set at the same time.

janedbal:
Thx, I still plan to do it, I just need to find some time :)

janedbal:
@kukulich I think it is reviewable. 

I was not sure what is the best approach for the restriction that both parameters cannot be set up together while using some default value to maintain BC. See the implementation, I think it should cause no issues like this.

kukulich:
Thanks!

Issue #1496, opened 2023-01-16
EarlyExit does not report useless else


momala454:
The issue reported in 2020 seems to still happen in 8.8.0
https://github.com/slevomat/coding-standard/issues/967

```php
if (1) {
            throw new \Exception('');
        } elseif (1) {
            echo 'bla';
        } else {
            echo 'hello';
        }

        if (1) {
            throw new \Exception('');
        } else {
            echo 'hello';
        }
```

Only the useless `elseif` is detected, not the two useless else

```
-----------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
-----------------------------------------------------------------------------------------
 94 | ERROR | [x] Use "if" instead of "elseif".
-----------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------
```

kukulich:
Please try to prepare PR with failing test. I'm not able to reproduce it.

momala454:
@kukulich , added a pull request https://github.com/slevomat/coding-standard/pull/1520

PR #1520, opened 2023-02-06
Add test for EarlyExit does not report useless else

momala454:
For issue https://github.com/slevomat/coding-standard/issues/1496


kukulich:
@momala454 So you expect it should report the `else` branch with comment?

momala454:
yes

kukulich:
> Removing "//" and keeping return false doesn't change the report, the else is not reported as useless

So please update the test to the relevant version.

momala454:
I updated the test

kukulich:
Thanks, fixed in https://github.com/slevomat/coding-standard/commit/fe768682a188034717b8d0f9411172af5a17417f

PR #1524, opened 2023-02-27
added rule for disallowing usage of variables inside strings

Ekimik:
Hello,
we are using your awesome coding standard rules inside our team own coding standard. We have large codebase, which runs on PHP 8.1, and where until recently no coding standard was used. We are slowly adopting our coding standard to our codebase, but it will be long way to do it everywhere. Soon we have plan to upgrade to PHP 8.2 and one of many deprecated features in PHP 8.2 is one type of [variables usage inside strings](https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated#dollar-outside). 
As the deperacated syntax is used in many places in our project, we would like to notify developers, that they should not use it in code, which is already plugged in CI code style check. 
Some developers may consider this kind of variables usage ugly, so there are configuration options, which allows rule users to define, which kinds of syntax should not be supported in their codebase.
It's simple checks based on regular expressions, which are not super smart and strict, but I believe it's good enough. Of course feel free to add notes and point to problems, which may occur.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,7 @@ Slevomat Coding Standard for [PHP_CodeSniffer](https://github.com/squizlabs/PHP_
  - [SlevomatCodingStandard.PHP.TypeCast](doc/php.md#slevomatcodingstandardphptypecast-) 🔧
  - [SlevomatCodingStandard.PHP.UselessParentheses](doc/php.md#slevomatcodingstandardphpuselessparentheses-) 🔧
  - [SlevomatCodingStandard.PHP.UselessSemicolon](doc/php.md#slevomatcodingstandardphpuselesssemicolon-) 🔧
+ - [SlevomatCodingStandard.Strings.DisallowVariableParsing](doc/strings.md#slevomatcodingstandardstringsdisallowvariableparsing)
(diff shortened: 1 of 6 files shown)


kukulich:
Thanks but there's `Squiz.Strings.DoubleQuoteUsage` sniff in PHPCS itself. Is there some benefit/improvement in your sniff?

Ekimik:
Thanks for pointing out to `Squiz.Strings.DoubleQuoteUsage`, I din't notice this sniff before. As far as I understand sniff `Squiz.Strings.DoubleQuoteUsage`, it disallows usage of double quoted strings, where it's not necessary and most importantly it disallows usage of any variable (no matter the syntax) inside double quoted string. Is that right?

What my sniff does in addition is
- it let's you choose, which syntax of writing variable inside string is ok for you (`$...`, `{$...}`). Syntax `${...}` is disallowed by default as it its deprecated as of PHP 8.2.
- it deals with heredoc as well, but I do not use this kind of string writing much, so I don't know exactly how people tend to use it with variables

kukulich:
Thanks!

Issue #1530, opened 2023-03-12
Namespaces.UnusedUses.UnusedUse removes despite being used in annotations in code, e.g. @var


kkmuffme:
The rule removes "use" statements, despite them being used in annotations inside the code, e.g. @var

Config:

```
<rule ref="SlevomatCodingStandard.Namespaces.UnusedUses">
	<properties>
		<property name="searchAnnotations" value="true"/>
	</properties>
</rule>
```

e.g. common when type guarding external libs/using static analysis:

```php
use Foo\Bar;

/**
 * @var Bar
 */
$foo = some_func();
```

or also a very common case in e.g. WordPress:

```php
use Foo\Bar;

/**
 * @param Bar|false
 */
$foo = apply_filters( 'hello', false );
```

or

```php
use Foo\Bar;

/**
 * @param Bar|false $arg
 */
$foo = apply_filters( 'hello', $arg );
```

or

```php
use Foo\Bar;

/**
 * @param int|false
 * @param Bar $arg
 */
$foo = apply_filters( 'hello', false, $arg );
```

kukulich:
Please provide a PR with failing test.

kukulich:
The bug is caused by https://github.com/phpstan/phpdoc-parser - there's no support for `@param` without parameter name.

kukulich:
Yes, it's probably reported as `InvalidTagValueNode`. However I don't want to add support for invalid tags. I use https://github.com/phpstan/phpdoc-parser so I don't have to do annotations parsing by myself.

TakeruDavis:
I would like to add to this a very similar problem with this rule, it doesn't work with generics in annotations either and removes the import

```
use Foo\Bar;

/**
 * @var \Collection<Bar>
 */
$foo = some_func();
```

kukulich:
@TakeruDavis Generics should work. Please try to prepare PR with failing test.

TakeruDavis:
@kukulich False alarm, looks like we've been locked on an older version of the package

PR #1531, opened 2023-03-12
Add failing tests for offlabel use of @var, @param 

kkmuffme:
by static analysis (psalm, phpstan, phpstorm) and WordPress
for Namespaces.UnusedUses.UnusedUse

https://github.com/slevomat/coding-standard/issues/1530


kukulich:
You added 5 use-cases but it looks only one of them is broken: `Failed asserting that 3 is identical to 2.` (one added error). What use-case is broken?

kukulich:
See https://github.com/slevomat/coding-standard/issues/1530#issuecomment-1465165815

PR #1542, opened 2023-03-28
Add array/DisallowPartiallyKeyedSniff

bkdotcom:
Array definitions must have keys specified for either all or none of the values


kukulich:
Thanks!

PR #1555, opened 2023-04-14
Do not sort if there is a Group Use

jonathan1055:
First just add test file to demonstrate the problem

### DIFF
--- a/tests/Sniffs/Namespaces/AlphabeticallySortedUsesSniffTest.php
+++ b/tests/Sniffs/Namespaces/AlphabeticallySortedUsesSniffTest.php
@@ -79,6 +79,19 @@ class AlphabeticallySortedUsesSniffTest extends TestCase
 		self::assertAllFixedInFile($report);
 	}
 
+	public function testIncorrectOrderGroupUseNotFixable(): void
+	{
+		$report = self::checkFile(
+			__DIR__ . '/data/alphabeticalGroupUseNotFixable.php',
+			[],
+			[AlphabeticallySortedUsesSniff::CODE_INCORRECT_ORDER]
+		);
+
+		self::assertSniffError($report, 5, AlphabeticallySortedUsesSniff::CODE_INCORRECT_ORDER, 'A');
+
+		self::assertAllFixedInFile($report);
+	}
+
 	public function testFixable(): void
 	{
 		$report = self::checkFile(
--- a/tests/Sniffs/Namespaces/data/alphabeticalGroupUseNotFixable.fixed.php
+++ b/tests/Sniffs/Namespaces/data/alphabeticalGroupUseNotFixable.fixed.php
@@ -1,0 +1,5 @@
+<?php
+
+use thirdproject\D;
+use otherproject\{B, C};
+use myproject\A;
--- a/tests/Sniffs/Namespaces/data/alphabeticalGroupUseNotFixable.php
+++ b/tests/Sniffs/Namespaces/data/alphabeticalGroupUseNotFixable.php
@@ -1,0 +1,5 @@
+<?php
+
+use thirdproject\D;
+use otherproject\{B, C};
+use myproject\A;


jonathan1055:
As expected, we get the test failure
```
1) SlevomatCodingStandard\Sniffs\Namespaces\AlphabeticallySortedUsesSniffTest::testIncorrectOrderGroupUseNotFixable
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
 '<?php\n
 \n
+use myproject\A;\n
 use thirdproject\D;\n
-use otherproject\{B, C};\n
-use myproject\A;\n
 '
```
which shows that group use statement `use otherproject\{B, C};` has been deleted during the fixer process.

SlevomatCodingStandard/Sniffs/Namespaces/AlphabeticallySortedUsesSniff.php
```diff
@@ -81,6 +81,11 @@ public function process(File $phpcsFile, $openTagPointer): void
 								break;
 							}
 						}
+						// Also if there are any 'group use' statements then we cannot sort and fix the file.
+						$groupUsePointer = TokenHelper::findNext($phpcsFile, T_OPEN_USE_GROUP, $openTagPointer);
+						if ($groupUsePointer) {
```
jonathan1055, line 86: I have fixed this now

kukulich:
Hmm, it looks there's problem that `UseStatementHelper::getFileUseStatements()` does not support group uses. This sniff is therefore not able to report wrong order for group uses. The best way is problably to move the `if` to the top of the `process` method and just ignore files with group uses.

jonathan1055:
I would say that if a project's standard includes the `SlevomatCodingStandard.Namespaces.AlphabeticallySortedUses` sniff then we _should_ report the incorrectly-ordered statements if we can detect that by looking at the single non-group statements, not silently ignore it just because we also detect a group use statement. The project standard may or may not include the `SlevomatCodingStandard.Namespaces.DisallowGroupUse` sniff, but in either case this sorting sniff should report that the single statements are out of order (as done in the PR), even if we can't fix it automatically. Then the developer can resolve it manually.

A future enhancement could be to include the group use statements when sorting (but that's not what I am suggesting to do now). But if we ignore the file completely then we are only doing half the job, and we are allowing source code files to pass standards that we know it should fail on.

kukulich:
So my solution in https://github.com/slevomat/coding-standard/pull/1555#issuecomment-1510428163 is a lot easier. We can also add comment to README that this sniff ignores group uses.

jonathan1055:
So you are saying that it is better not to report anything even when we know that the single use statements are out of order? The only downside of doing it the way I have coded is that there will be cases when we report nothing because the single statements are in order but a group statement may be present somewhere which we've ignored when checking the order. Is it not better to report a definite problem when we find it, even if there are other cases that don't get reported?

I will move the `if` statement if that is what you want, but I just want to make sure I understand the philosophy behind the decision.

kukulich:
> we know that the single use statements are out of order

We don't know it "right" because the parsing does not support it in the right way.

jonathan1055:
> We don't know it "right" because the parsing does not support it in the right way.

I do not understand what you mean. Yes, the group statements are dropped, and only the single statements are kept. But if we find a sort order error in those single use statements then we do know that the overall order is wrong, regardless of whether any other group statements are present, because at minimum there is a sort order problem in the single statements. 

Yes, we do not know the whole truth. But if we find one falsehood we can say for sure that overall it has to be false. I'm not trying to argue and be a nuisance :-) I just want to understand the logic behind your position. Thank you for taking the time and trouble to have this discussion, I appreciate it.

jonathan1055:
What is the proper workflow required here? I synced my own repo with slevomat/coding-standards, and that seemed fine and clean. But the PR pipeline stopped with "merge commit blocked". So I rebased locally to remove the merge commit and force-pushed back to my repo. But now, all those other commits are showing in the PR above. Clearly this is not what you want. What is your desired workflow in these situations?

RMEngels:
@jonathan1055 you could try `git rebase -i HEAD~n` where n is the number of commits you want to rebase off of `HEAD` (I guess 26 in this case) and put `DROP` in front of the ones you don't want anymore. After the rebase force push to your branch

jonathan1055:
Thanks @RMEngels I have rebased to remove the 23 unwanted commits, leaving just the 3 that should be in this PR. This is fine, the PR is clean but is out of sync with the master branch. This is exactly where I was before my comment on 15 May above. I used "sync fork" on my own github repo last time. Maybe I will try the "Update branch" from here instead, not my own repo.

jonathan1055:
Yes after my comment above I used the "update branch - with rebase" from the dropdown in this PR and it is now clean here.

abelapko:
I also get this bug. This fix is important.

jonathan1055:
The only check to fail was Code Coverage. Looking at the details on https://app.codecov.io/gh/slevomat/coding-standard/commit/1f223f11ffb4b56debca3c3187b7279be3846eee it implies that the one file directly changed in this PR is 100% covered, but three files with "indirect" changes have reduced coverage. I'm not sure what this means though.

kukulich:
The PR won't be merged in this version anyway.

jonathan1055:
OK, I will make the change that you wanted.
The code coverage report is an excellent tool.

jonathan1055:
Thanks for the merge.

PR #1575, opened 2023-05-19
TestCase: generate sniff name fix with CodeSniffer method

JanTvrdik:
PHP_CodeSniffer's logic includes a bit more magic. This makes the behavior consistent.


kukulich:
Thanks!

PR #1597, opened 2023-07-12
Fix TypeError via JumpStatementsSpacingSniff when return is on the first line of the file

herndlm:
This fixes
```
TypeError: SlevomatCodingStandard\Helpers\FixerHelper::removeBetweenIncluding(): Argument #3 ($endPointer) must be of type int, null given, called in <redacted>/coding-standard/SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacing.php on line 221
```

`$endOfLineBeforePointer` can be `null`, but `FixerHelper::removeBetweenIncluding()` expects `int`.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacing.php
+++ b/SlevomatCodingStandard/Sniffs/ControlStructures/AbstractControlStructureSpacing.php
@@ -217,7 +217,9 @@ abstract class AbstractControlStructureSpacing implements Sniff
 			$phpcsFile->fixer->replaceToken($pointerBefore, '<?php');
 		}
 
-		FixerHelper::removeBetweenIncluding($phpcsFile, $pointerBefore + 1, $endOfLineBeforePointer);
+		if ($endOfLineBeforePointer !== null) {
+			FixerHelper::removeBetweenIncluding($phpcsFile, $pointerBefore + 1, $endOfLineBeforePointer);
+		}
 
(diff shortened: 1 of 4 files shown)


kukulich:
Thanks.

herndlm:
unrelated, but I got another weird one you might be able to help me with

```
PHP Fatal error:  Uncaught TypeError: SlevomatCodingStandard\Helpers\AnnotationHelper::getStartPointer(): Return value must be of type int, null returned in <redacted>/vendor/slevomat/coding-standard/SlevomatCodingStandard/Helpers/AnnotationHelper.php:326
```

The code it wants to parse seems to be
```php
    /**
     * @param string|array<string, string> $atts
     *
     */
    public function render(string|array $atts, string|null $content = null): string
```

image

Line 24 is that @param. any ideas? I don't know the code here well enough :)

kukulich:
@herndlm Are you able to prepare failing test?

herndlm:
I showed the wrong PHPDoc apparently.. weirdly I cannot reproduce it in a test yet. But so far I know that it is triggered via `UnusedUsesSniff` with `searchAnnotations` enabled.

In our app it crashed with
```php
<?php

/**
 * Created by PhpStorm.
 * User: herndlm
 * Date: 11/04/1990
 * Time: 12:34
 */

namespace Foo;
```
fixed by removing the phpdoc.

another crash happened with
```php
<?php defined('ABSPATH') or exit(403);

/**
 * @var array $cssClasses;
 * @var string $content;
 */
?>

<table class="<?php echo implode(' ', $cssClasses);?>">
    <?php echo do_shortcode($content);?>
</table>
```
again fixed by removing the phpdoc

PR #1622, opened 2023-09-15
Don't treat a group use as a constant

### LINKED ISSUE #1621: Bug in `SlevomatCodingStandard.Namespaces.FullyQualifiedGlobalConstants`
I found a bug when using these rules:
```xml
<rule ref="SlevomatCodingStandard.Namespaces.FullyQualifiedGlobalConstants" />
<rule ref="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly">
    <properties>
        <property name="searchAnnotations" value="true" />
        <property name="allowFullyQualifiedGlobalConstants" value="true" />
    </properties>
</rule>
```

In the following code, `User` and `Role` will be formatted to `\User` and `\Role` (See image below).
```php
<?php

namespace App;

use App\Entity\{ User, Role as EntityRole};

final class Foobar
{
    /**
     * @var User
     */
    private $user;

    /**
     * @var EntityRole
     */
    private $role;
}
```

**:information_source:  NOTE:**
- I already use `SlevomatCodingStandard.Namespaces.DisallowGroupUse` to prevent group use.

asispts:
This PR will fix #1621

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Namespaces/AbstractFullyQualifiedGlobalReference.php
+++ b/SlevomatCodingStandard/Sniffs/Namespaces/AbstractFullyQualifiedGlobalReference.php
@@ -17,6 +17,7 @@ use function array_map;
 use function sprintf;
 use function strtolower;
 use const T_OPEN_TAG;
+use const T_OPEN_USE_GROUP;
 
 /**
  * @internal
@@ -61,6 +62,10 @@ abstract class AbstractFullyQualifiedGlobalReference implements Sniff
 	public function process(File $phpcsFile, $openTagPointer): void
 	{
 		if (TokenHelper::findPrevious($phpcsFile, T_OPEN_TAG, $openTagPointer - 1) !== null) {
+			return;
+		}
+
+		if (TokenHelper::findNext($phpcsFile, T_OPEN_USE_GROUP, $openTagPointer) !== null) {
 			return;
 		}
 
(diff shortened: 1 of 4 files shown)


SlevomatCodingStandard/Sniffs/Namespaces/AbstractFullyQualifiedGlobalReference.php
```diff
@@ -75,6 +76,11 @@ public function process(File $phpcsFile, $openTagPointer): void
 			$name = $referencedName->getNameAsReferencedInFile();
 			$namePointer = $referencedName->getStartPointer();
 
+			// We are in the middle of group use declaration
+			if (TokenHelper::findNextLocal($phpcsFile, T_CLOSE_USE_GROUP, $namePointer) !== null) {
```
kukulich, line 80: I don't this is right. I think we should just ignore the complete file if there's any group use. 

We don't support anything about group uses, so it's dangerous to try to fix the file.

kukulich (changes requested):

SlevomatCodingStandard/Sniffs/Namespaces/AbstractFullyQualifiedGlobalReference.php
asispts: Ok, done.

kukulich (approved):

kukulich:
Thanks!

PR #1623, opened 2023-09-28
Arrays/DisallowImplicitArrayCreation: bug fix - recognize `global` statements

jrfnl:
While it may be unclear whether the variable imported via the `global` statement is an array, the variable does exist, so this is not implicit array creation.

Includes tests.


kukulich:
Thanks.

PR #1627, opened 2023-09-28
NamespaceHelper::getAllNamespacesPointers(): bug fix - allow for namespace tokens used as operator

jrfnl:
The `namespace` keyword can be used to declare namespaces, but is also allowed as an operator which resolves to the current namespace: `namespace\callMe();`.

This was not taken into account correctly when gathering all namespace declaration pointers.

Discovered via a false positive received from the `Namespaces/UnusedUses` sniff.

Includes tests for both the `NamespaceHelper` method as well as the `Namespaces/UnusedUses` sniff.


kukulich:
Thanks.

PR #1629, opened 2023-10-07
Fix more typos

szepeviktor:
Three functions have misspellings.

PR #1630, opened 2023-10-07
Run Typos in CI

szepeviktor:
Needs #1629 to succeed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -208,3 +208,15 @@ jobs:
           files: temp/coverage.xml
           flags: unittests
           fail_ci_if_error: true
+
+  typos-check:
+    name: "Typos"
+
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+
+      - name: "Search for misspellings"
+        uses: "crate-ci/typos@v1.16.17"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -208,3 +208,15 @@ jobs:
...
+    name: "Typos"
+
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+
+      - name: "Search for misspellings"
+        uses: "crate-ci/typos@master"
```
kukulich, line 222: Is it possible to use v1 or something like this?

szepeviktor: No, they don't have this kind of tag.
https://github.com/crate-ci/typos/tags
Typos is getting better at typos fast.

BTW I find their GHA action terrible.
Using this instead https://github.com/szepeviktor/byte-level-care/blob/f1c750a0424ab8a0b7e96cd253f8cb3973b324c6/.github/workflows/spelling.yml#L31-L50

kukulich: Ok, so I would use the latest tag v1.16.17

szepeviktor: Done 🍏

kukulich:
Thanks.

szepeviktor:
Glad to meet your bytes.

PR #1661, opened 2024-02-06
Fix detection of class usage in double-quoted strings and heredoc

c01l:
I encountered a bug in `Sniffs.Namespaces.UnusedUsesSniff` that resulted in reporting too many "unused" classes, if a class instantiation occurred inside of a double quoted string or heredoc.

Quick example (HTML composition):

```php
<?php

use SomeNamespace\MessageBox;

$_ = fn($x) => strval($x);
$html .= <<<HTML
<div>
   <div>...</div>
   {$_(new MessageBox("User did something wrong."))}
</div>
HTML
```

```php
<?php

namespace SomeNamespace;

class MessageBox {
  public function __construct(private string $msg) {}
  public function __toString(): string {
    return "<div class='alert alert-danger'>$this->msg</div>";
  }
}
```

I added some tests for this case and added logic for detecting T_NEW T_STRING as class references.


kukulich:
Thanks.

PR #1666, opened 2024-02-19
issue: DocCommentSpacingSniff - same line

michalbundyra:
When we have two PHPDocs in a single line we are getting an error:

```console
| ERROR | [x] Expected 1 line between annotations groups, found -1. (SlevomatCodingStandard.Commenting.DocCommentSpacing.IncorrectLinesCountBetweenAnnotationsGroups)
```

but it is not possible to fix it automatically.

And I am not sure how to fix in in the code atm, so just providing the failing example... Thanks!


kukulich:
The main problem is that PHPCS itself cannot parse the tags on same line - PHPCS things it is only one tag with description `self::method() @link https://example.com`. Please report the bug to PHPCS first.

Issue #1471, opened 2022-11-23
`SlevomatCodingStandard.Classes.ClassStructure` rule, __invoke method


zlodes:
Hi!

Would be nice to have `__invoke` as independent group. Config example:
```xml
<element value="constructor"/>
<element value="invoke"/>
<element value="all public methods"/>
<element value="methods"/>
<element value="magic methods"/>
``` 

Code example:
```php
final class Controller
{
    public function __construct()
    {
    }

    public function __invoke()
    {
        $this->doSomething();
    }

    private function doSomething(): void
    {
    }
}
```

Seems there is a similar issue: https://github.com/slevomat/coding-standard/issues/1422,
but this seems easier to implement.

dpi:
I think #1098 captures what we want to achieve here.

Issue #1098, opened 2020-10-12
ClassStructureSniff individual methods configuration


maryo:
Hello. I'd like to have my PHPUnit `protected function setUp` method placed before public test methods. But it looks like it is not possible to configure something like an exception or a custom group or something like that. Do you want to support such a feature? I made them public as a workaround.

Thanks

pscheit-lillydoo:
How would this be implemented? The ability to match several methods via regexp into a custom group, that one can then order in between the builtin groups?

If someone could draft a configuration, I would give it a try to implement

nederdirk:
I'd also like something like this, could a (phpdoc) attribute like `#[SlevomatCodingStandard\ClassStructure\Group('myCustomGroupName')]` (and/or comparable PHPdoc tag) work to (optionally) override the automatically-determined group?

Something like (taken from a hypothetical datetime-library):
```php
#[SlevomatCodingStandard\ClassStructure\Group('iso8601')]
public function toIso8601(): string
{
//should sort with other 'iso8601' tagged functions iff the 'iso8601' group is in the configuration, otherwise it should sort with the 'public methods'
}

#[SlevomatCodingStandard\ClassStructure\Group('iso8601')]
public static function fromIso8601(string $timestamp): DateTime
{
//should sort with other 'iso8601' tagged functions iff the 'iso8601' group is in the configuration, otherwise it should sort with the 'public static constructors'
}
```

vhenzl:
It would also be good to be able to order/group methods by a specified PhpDoc tag (e.g. `@api`).

saas786:
We face same issue.

```
class MyClass {
    public function __construct() {
        $this->boot();
    }

    private function boot() {}

}
```
OR
```
class MyClass {
    public function __construct() {
        $this->init();
    }

    private function init() {}

}
```

Either of these cases, I would like my `boot()` or `init()` to show up after `__construct()` but `ClassStructureSniff ` moves it below other functions, sometimes at bottom of classes, which is not ideal. makes it hard to navigate the class.

would be nice if I could do something like:

```
				<element value="constructor"/>
				<element value="custom_boot"/>
				or
				<element value="custom_init"/>
```

PR #1683, opened 2024-05-27
`SlevomatCodingStandard.Classes.ClassStructure`: Group for __invoke method

freerich:
As mentioned in issue #1471, we need to give special treatment to __invoke methods. In some applications, __invoke methods are as important as __construct methods, such as in EventHandler, CommandHandler, and invokable controllers.

Adding this specific “invoke” case might be much easier than handling individual method names, as discussed in issue #1098.


kukulich:
Thanks I've implemented it here: https://github.com/slevomat/coding-standard/commit/ad015190c736a8878c25fb881a6a3ee41c4f2382

PR #1693, opened 2024-08-19
DeclareStrictTypesSniff: Fixing number of empty lines after line comment

maryo:
When the previous effective token before declare statement is line comment (`// ...`/`# ..`), the comment spans across the entire line and there is no newline whitespace token. The number of lines was counter incorrectly in such case.


kukulich:
Thanks.

PR #1700, opened 2024-10-07
Fix order of sections

tfrommen:
👋 Hi there!

The diff looks a little wild, but I essentially only swapped the sections for the `UseFromSameNamespace` and the `UseDoesNotStartWithBackslash` sniffs, which were not in the correct alphabetically sorted order.


kukulich:
Thanks.

PR #1702, opened 2024-10-14
Create a sniff to ensure constants being typed by native types

### LINKED ISSUE #1701: New sniff to ensure constants being typed using native types
Currently there's no sniff to ensure that constants are typed through native types.

I'd like to introduce a sniff that enables this and ensures the following:
```php
class A
{

	const null AA = null;
	const true AAA = true;
	const false AAAA = false;
	const string AAAAA = 'aa';
	const int AAAAAA = 123;
	const float AAAAAAA = 123.456;
	const array AAAAAAAA = ['php'];
	const int AAAAAAAAA = -123;
	const float AAAAAAAAAA = -123.456;

}
```

This is available since PHP 8.3. See: https://wiki.php.net/rfc/typed_class_constants

DaDeather:
This resolves #1701

### DIFF
--- a/SlevomatCodingStandard/Sniffs/TypeHints/MissingNativeConstantTypeSniff.php
+++ b/SlevomatCodingStandard/Sniffs/TypeHints/MissingNativeConstantTypeSniff.php
@@ -1,0 +1,144 @@
+<?php declare(strict_types = 1);
+
+namespace SlevomatCodingStandard\Sniffs\TypeHints;
+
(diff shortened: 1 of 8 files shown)


kukulich:
Many thanks to you! However I've implemented it myself because I've combined it with one already existing sniff: https://github.com/slevomat/coding-standard/commit/b022231e75f14598143333112f935a5bcf6cebd4

PR #1704, opened 2024-10-21
Add support for allowing public readonly properties

tfrommen:
In general, public properties are not great, and one should use method access where possible.

However, read-only properties are not that bad, which is why this PR adds support for allowing them via a new public property (ha-ha) on the `ForbiddenPublicProperty` sniff, in a backwards-compatible way.


tfrommen:
Hi @kukulich,

I was wondering if there is an interest in allowing for `public readonly` properties...?
If there is, I'm happy to get feedback on my approach. If not, I can easily use a custom sniff (maybe based on the Slevomat one, or not).

Thanks!

kukulich:
@tfrommen Yes, this PR will be accepted :) I hope I'll prepare new CS version around the release of PHP 8.4.

kukulich:
Thank!

PR #1705, opened 2024-11-02
Helper/TokenHelper: remove use of T_ARRAY_HINT

jrfnl:
This token was deprecated in PHPCS 3.3.0 and has been unused in PHPCS since, so no need to look for it as the minimum PHPCS version for this standard is PHPCS 3.9.0.


kukulich:
Thanks.

PR #1708, opened 2024-11-07
Update to phpstan/phpdoc-parser 2.0

ondrejmirtes:
See the upgrading guide: https://github.com/phpstan/phpdoc-parser/blob/2.0.x/UPGRADING.md


umpirsky:
Can we get a new version with this change released please?

PR #1729, opened 2025-02-24
Fix reference to SlevomatCodingStandard.Arrays.ArrayAccess in README

jnoordsij:
- Move to its alphabetical place
- Fixup name to match sniff rather than PHP filename


kukulich:
Thanks.

PR #1736, opened 2025-03-24
TypeHints/DisallowMixedTypeHint: fix logic error

jrfnl:
If a file only contains a placeholder docblock (typically an empty `index.php` file in a OS project which doesn't have control over server configuration), it is possible for the `TokenHelper::findNextEffective()` method to return `null`.

As the code didn't take this into account, this led to the following error:
```
 1 | ERROR | [ ] An error occurred during processing; checking has been aborted. The error message was: Undefined array
   |       |     key "" in
   |       |     path/to/SlevomatCodingStandard/Sniffs/TypeHints/DisallowMixedTypeHintSniff.php on
   |       |     line 86
   |       |     The error originated in the SlevomatCodingStandard.TypeHints.DisallowMixedTypeHint sniff on line 86.
   |       |     (Internal.Exception)
```


kukulich:
Thanks!

Issue #1737, opened 2025-03-24
`SlevomatCodingStandard.Classes.ClassMemberSpacing` false negative when multiple attributes are used


janedbal:
Given following file

```php
<?php 

class Test
{

    public function method1(): void
    {
    }
    #[FirstAttribute]
    public function method2(): void
    {
    }
    #[FirstAttribute]
    #[AnotherAttribute]
    public function method3(): void
    {
    }

}
```

I'm getting error only for `method1`, but not `method2`:

```
Line 6: Expected 1 blank line after method, found 0.
```

Removing one of the attributes fixes the issue.

janedbal:
Looks like `SlevomatCodingStandard.Classes.MethodSpacing.IncorrectLinesCountBetweenMethods` has the same issue

kukulich:
I'm not able to repreduce it: https://github.com/slevomat/coding-standard/commit/8bf0408a9cf30687d87957d364de9a3d5d00d948

Could you please provide a failing test?

janedbal:
I think we can close this, [MethodSpacing is fixed](https://github.com/slevomat/coding-standard/commit/374ad13f74ac98df7f4c0781866bf4ac78ddeb01) and ClassMemberSpacing was not reproduced. If I ever encounter this problem again, I'll send failing test right away.

PR #1740, opened 2025-03-28
Failing test for MethodSpacingSniff with multiple attributes

janedbal:
Ref: https://github.com/slevomat/coding-standard/issues/1737

### DIFF
--- a/tests/Sniffs/Classes/MethodSpacingSniffTest.php
+++ b/tests/Sniffs/Classes/MethodSpacingSniffTest.php
@@ -17,7 +17,7 @@ class MethodSpacingSniffTest extends TestCase
 	{
 		$report = self::checkFile(__DIR__ . '/data/methodSpacingErrors.php');
 
-		self::assertSame(6, $report->getErrorCount());
+		self::assertSame(7, $report->getErrorCount());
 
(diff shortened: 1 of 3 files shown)


kukulich:
@janedbal I think there is a bit of confusion here about what sniff does. It doesn't check for spaces between individual methods, but between all methods and all properties, constants etc..

janedbal:
I'm truely confused. Lets point to docs then:

> SlevomatCodingStandard.Classes.MethodSpacing: Checks that there is a certain number of **blank lines between methods**.
> SlevomatCodingStandard.Classes.ClassMemberSpacing: Sniff checks lines count **between different class members**, e.g. between last property and first method.

Here I sent failing test for MethodSpacingSniff. So that one is not checking spaces between individual methods? Maybe docs needs some adjustment.

kukulich:
@janedbal I'll happy approve PR with better documentation :) And I will check this report.

janedbal:
If this failing test is valid, I think docs are fine.

kukulich:
Thank you for your report.

Fixed in https://github.com/slevomat/coding-standard/commit/374ad13f74ac98df7f4c0781866bf4ac78ddeb01

PR #1748, opened 2025-04-29
RequireMultiLineCall: Fix when multiline call exactly fits into the line length limit

maryo:
Fixing reporting error for a multiline call that exactly fits into the line length limit.

```php
$array = array_map(['a loooooooooooooooooong value that exactly fits into the line length limit'], function (): string {
    return 'foo';
});
```

Here, `$lineStart` would be `        $array = array_map(`.
Previously, `$lineEnd` was `(['a loooooooooooooooooong value that exactly fits into the line length limit'], function (): string {`.
It was including the opening parenthesis `(` again, so `$lineLength` of this line was `121` instead of `120`.


kukulich:
Thanks.

PR #1749, opened 2025-04-29
SlevomatCodingStandard.Classes.ClassStructure: group based on method name prefix

maryo:
Support for custom method group based on method name prefix.


kukulich:
Thanks.

PR #1750, opened 2025-05-07
InlineDocCommentDeclarationSniff: fix error when no code is found after @var

AegirLeet:
Currently, with `allowDocCommentAboveReturn` set to `true`, the sniff errors with `Undefined array key "" in InlineDocCommentDeclarationSniff.php on line 123` when a `@var` annotation is not followed by any other code, because `$pointerAfterCommentClosePointer` ends up being `null` in that case.


kukulich:
Thanks.

PR #1755, opened 2025-05-23
SlevomatCodingStandard.TypeHints.ClassConstantTypeHint: fixing only private constants

maryo:
It is dangerous to automatically add typehints for other than private constants as the type might not be compatible with constants in the inheritance chain or implemented interfaces.

Or if you prefered this to be configurable, I could make it configurable.


maryo:
OK, I added `fixableNativeTypeHint`: `yes`/`no`/`private`. I used `yes`/`no` simmilar to `SlevomatCodingStandard.TypeHints.DNFTypeHintFormat` so all possibilities are strings instead of `string|bool`
 but I can change it if you prefer different configuration values or name.

PR #1756, opened 2025-05-25
SlevomatCodingStandard.Classes.ClassMemberSpacing: Prevent deleting code

maryo:
Given this code
```php
private int $property;


/**
 * Description
 */

public function method(): void
{
}
```

The sniff deleted the docblock entirely because there is (quite unexpected, still a valid docblock though) newline after it.

Also, there could be some extra comments like this:

```php
private int $property;

// Extra comment

/**
 * Description
 */
public function method(): void
{
}
```

I'm not sure what the ideal solution would be here. I changed it so that the extra comment is considered part of the method.

But there could even be some unexpected code (it would be probably invalid though) which should not get deleted either. In such case, i disabled the fix and just report it as an error.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/ClassMemberSpacingSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/ClassMemberSpacingSniff.php
@@ -63,7 +63,6 @@ class ClassMemberSpacingSniff implements Sniff
 		$tokens = $phpcsFile->getTokens();
 
 		$memberPointer = null;
-		$previousMemberPointer = null;
 
 		do {
 			$previousMemberPointer = $memberPointer;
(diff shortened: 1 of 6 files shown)


SlevomatCodingStandard/Sniffs/Classes/ClassMemberSpacingSniff.php
```diff
@@ -240,7 +247,11 @@ private function getMemberFirstCodePointer(File $phpcsFile, int $memberPointer):
 			/** @var int $firstCodePointer */
 			$firstCodePointer = TokenHelper::findPrevious(
 				$phpcsFile,
-				[T_VAR, T_PUBLIC, T_PROTECTED, T_PRIVATE, T_ABSTRACT, T_FINAL, T_SEMICOLON, T_CLOSE_CURLY_BRACKET],
+				array_merge(
```
maryo, line 250: `T_STATIC` and `T_READONLY` were missing.

kukulich:
Thanks.

PR #1761, opened 2025-06-09
SlevomatCodingStandard.Classes.PropertyDeclaration: Prevent confusion when type hint is missing

vojtech-dobes:
I've noticed that when property doesn't have native type hint, `PropertyDeclaration` gets confused if the class also has property-like modifiers (eg. `final`). It led to unintended screw up when auto-fixing.

For example this class https://github.com/vojtech-dobes/php-graphql-server/blob/7f0587879ed4ce58d866bbeee2c301d98aca0c03/src/GraphQL/Deferred.php#L11-L15 got auto-fixed to:

```php
final class Deferred {

	/** @var callable(): TValue */
	private $callback;
```

I don't know exactly how to create test for this.


kukulich:
Thanks. Should be fixed in 8.19.1.

PR #1762, opened 2025-06-17
Allow targeting multiline elements in AbstractPropertyConstantAndEnumCaseSpacing

vojtech-dobes:
Hi, feel free to reject this if you don't like this. I love the nice spacing I can define with the various Spacing sniffs. I use 1 empty line above phpDoc and 0 lines otherwise. To me it's about visually putting apart more complex elements (with phpDoc a property basically becomes multiline even if it's just single-line on its own). But I noticed that multiline `const` are still crammed together. The reason is that they don't really need phpDoc (eg. PHPStan can read them precisely just fine without phpDoc).

I've tried to implement these 2 extra options `minLinesCountBeforeMultiline` & `maxLinesCountBeforeMultiline` as optional. By default they are off (value `-1`). They also still respect other rules, so most _demanding_ value always wins.

The `max` call on line 105 I'm not sure about. I noticed weird behavior when the min is higher than max, so this ensures max is always same or higher than min.


vojtech-dobes:
Thank you!

PR #1764, opened 2025-06-19
ClassHelper::isFinal() - param 2 must be an int... null passed

bkdotcom:
Got an error regarding ClassHelper::isFinal requiring an int, `null` passed

quick and easy fix... sorry, no unit tests


kukulich:
@bkdotcom Do you have code that throws this error?

kukulich:
Thanks, I've fixed it this way: https://github.com/slevomat/coding-standard/commit/7f629592aa2efde9fd7c158b57517796fbb00b3e

PR #1765, opened 2025-06-30
Allowed null in TraitUseSpacingSniff.linesCountAfterLastUse

vrana:
We have other rules, namely `FunctionSpacingSniff.spacingBeforeFirst` which collide with this.


kukulich:
Thanks!

PR #1770, opened 2025-08-10
SlevomatCodingStandard.Classes.RequireMultiLineMethodSignature: New option "withPromotedProperties"

vrana:
A similar option should be probably in `RequireSingleLineMethodSignature`. However, `minParametersCount` is not there either. I think that the best solution would be to merge these two into one but it seems quite complicated mainly because of BC. If you agree, I'll add `withoutPromotedProperties` to `RequireSingleLineMethodSignature`.


kukulich:
May I suggest something like “forceInConstructor” ?

vrana:
I don't want it to trigger in all constructors, only in those with promoted properties.

kukulich:
Oh, you’re right

vrana:
Can I get this in?

kukulich:
I will check and merge it probably in the weekend after vacation.

Issue #1074, opened 2020-08-21
Suggestion: whitelist for unused variables/parameters


flaviovs:
It would be good if unused variables and parameters sniffs supported a whitelist of names, because PHPCS suppression is not granular enough to always flag only the variables that are unused intentionally. This is common in other languages (e.g. Python's `_`)

Consider the following example:

```php
function onSomething($ev, $foo) {
    echo $foo;
}
```

Now `$ev` will generate a _SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter_. To suppress the error one can do this:

```php
// phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter
function onSomething($ev, $foo) {
    echo $foo;
}
```

Now a few months later somebody changes the code to:
```php
// phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
function onSomething($ev, $foo) {
    echo 'event triggered: ' . $ev;
}
```

Now the sniff will never catch the unused parameter `$foo`, unless the programmer remember to remove the suppression (unlikely, in my experience).

Same for variables:

```php
[$h, $m, $s] = explode(':', $time_hms);
// (...many lines of code...)
echo $h;
// Unused variables $m and $s
```
Someone suppress the unused variable error:

```php
// phpcs:ignore SlevomatCodingStandard.Variables.UnusedVariable
[$h, $m, $s] = explode(':', $time_hms);
// (...many lines of code...)
echo $h;
// All good.
```

Now someone decides to implement it differently:

```php
// phpcs:ignore SlevomatCodingStandard.Variables.UnusedVariable
[$h, $m, $s] = explode(':', $time_hms);
// (...many lines of code...)
echo get_hms_hour($time_hms);
// $h, $m, and $s now unused forever.
```

This suggestion entails adding a property to the sniffs:

```xml
<!-- Note: ideally it should do regex matching -->
<property name="whitelist" type="array" value="/^unused_.+/" />
```

Then one could write code like this:

```php
function onSomething($unused_ev, $foo) {
    echo $foo;

    [$h, $unused_m, $unused_s] = explode(':', $time_hms);

    echo $h;
}
```

This not only would solve PHPCS lack of granularity, but would make the code self-documenting and much more readable, IMHO.

srigi:
I would love to see some sort of pattern configuration implemented too. I like how [ESLint is handling](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) this case.

[...]

PR #1781, opened 2025-08-29
feat(UnusedParameterSniff): support supression via parameter name

claytonrcarter:
Hi 👋 

It's common to have unused method params, especially in cases where you're extending and overriding a method from a parent class. Instead of sprinkling `// phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter` thoughout my code, I'd prefer to just rename those variables to `$_variable`. This is similar to how eslint and Rust's `clippy` handle unused params, and I've found those to be convenient and still helpful.

My only TODO is maybe putting this behind a config flag, but I didn't want to get into that until I heard if this change is desired at all. This is my first contribution here, and no AI was used in this. 

This sort of addresses #1074, and it may even close it.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Functions/UnusedParameterSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Functions/UnusedParameterSniff.php
@@ -2,16 +2,20 @@
 
 namespace SlevomatCodingStandard\Sniffs\Functions;
 
+use Exception;
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
 use PHP_CodeSniffer\Util\Tokens;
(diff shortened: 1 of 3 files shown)


tests/Sniffs/Functions/data/unusedParameterNoErrors.php
```diff
@@ -29,6 +29,11 @@ function ($closureSuppress)
 
 };
 
+function usedParameterSupressedViaName($_)
```
claytonrcarter, line 32: TODO add another param to confirm that `$_` isn't magical and that `$_foo` is also allowed.

kukulich:
I'm ok to merge such feature but I would prefer to make it configurable via new option. The option should probably contain regexp.

PR #1783, opened 2025-09-05
ReturnTypeHint: report missing type hint in abstract methods

vrana:
Abstract methods now report fixable error _Method … does not have void return type hint_ which is wrong.


kukulich:
Thanks.

PR #1787, opened 2025-09-16
PHPCS 4.0

kukulich:


composer.json
```diff
@@ -16,7 +16,7 @@
 		"php": "^7.4 || ^8.0",
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.6.2 || ^0.7 || ^1.1.2",
 		"phpstan/phpdoc-parser": "^2.3.0",
-		"squizlabs/php_codesniffer": "^3.13.4"
+		"squizlabs/php_codesniffer": "4.0.0"
```
Xerkus, line 19: Was it meant to be the exact version rather than range?

PR #1791, opened 2025-09-18
Introduce rule RequireOneDocCommentSniff

kamil-zacek:


kukulich:
Please fix the missing test.

PR #1798, opened 2025-09-25
feat: suffix filtering in methodGroups of ClassStructureSniff

### LINKED ISSUE #1797: Allow "end with" method name in SlevomatCodingStandard.Classes.ClassStructure methodGroups
Actually, we can specify "start with" patthern in methodGroups of the rule SlevomatCodingStandard.Classes.ClassStructure.

It will be interesting to allow "end with" too. Our use case is for PHPUnit's data provider. Theses methods are "public static" and ends by "DataProvider".

In our standard, all "public static" methods need to be defined before "public" ones, except for data providers one which needs to be defined at the end of the file.

DjLeChuck:
closes #1797


kukulich:
Can you please also update the documentation? https://github.com/slevomat/coding-standard/blob/master/doc/classes.md

PR #1831, opened 2026-02-12
Sort import traits inside class

h4kuna:
Hi, I’ve prepared a sniff that sorts use statements inside a class alphabetically. Please take a look — I hope you’ll like it and that it can be merged into master. This is my first sniff.

Example:

```php
class UnsortedOneline
{
	use C, A, B;
}
```

fix
```php
class UnsortedOneline
{
	use A, B, C;
}
```


kukulich:
Looks good. Please:

- fix build
- add tests with combination of different type of use, eg. `use A, B\C, \D\E\F;`

h4kuna:
I added tests with combination of different type of use, the first `\` is ignored for the sort.

PR #1834, opened 2026-02-13
Sort throws annotations

h4kuna:
Hi, I prepared new sniff for sort @throws annotations in phpdoc alphabeticaly.

Example:
```php
class ErrorsClass
{
	/**
	 * @throws \InvalidArgumentException
	 * @throws \ArgumentCountError
	 * @throws \Exception
	 */
	public function unorderedThrows(): void
	{
	}
}
```
fix to
```php
class ErrorsClass
{
	/**
	 * @throws \ArgumentCountError
	 * @throws \Exception
	 * @throws \InvalidArgumentException
	 */
	public function unorderedThrows(): void
	{
	}
}
```


kukulich:
Thanks!

PR #1848, opened 2026-04-10
Fix ReferenceUsedNamesOnlySniff to use existing alias when colliding class names are allowed

Khartir:
When `allowFullyQualifiedNameForCollidingClasses` was enabled and a FQN reference collided on the unqualified name with another class, the sniff incorrectly skipped the reference even when a use statement with an alias
already existed for that exact FQN. Now the sniff detects the existing alias and flags the FQN for replacement in both the detection and fix phases.


kukulich:
Thanks!

PR #1849, opened 2026-04-16
Add full docblock support for DNFTypeHintFormatSniff

alleknalle:
Hi there!

This PR extends ``DNFTypeHintFormatSniff`` to enforce type hint formatting in docblocks

``DNFTypeHintFormatSniff`` previously only enforced its formatting rules on PHP code type hints. Docblock annotations (``@param``, ``@return``, ``@var``, ``@property``, ``@property-read``) were silently ignored, leaving them out of sync with the configured rules.

This PR extends the sniff to apply the same four rules to docblock type hints:
- ``withSpacesAroundOperators`` — spaces around | and & in union/intersection types
- ``withSpacesInsideParentheses`` — spaces inside ( ) in DNF intersection types
- ``shortNullable`` — ?Type vs Type|null (both directions)
- ``nullPosition`` — null first or last in union types (both directions)

All violations are auto-fixable, consistent with the existing PHP code behaviour.

These changes make the ``NullTypeHintOnLastPositionSniff`` redundant. ``NullTypeHintOnLastPositionSniff`` only ever applied to docblocks and is now fully covered by ``DNFTypeHintFormatSniff`` with ``nullPosition: 'last'``, which handles both PHP code and docblocks. I marked ``NullTypeHintOnLastPositionSniff`` ``@deprecated`` with a pointer to the replacement.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/TypeHints/DNFTypeHintFormatSniff.php
+++ b/SlevomatCodingStandard/Sniffs/TypeHints/DNFTypeHintFormatSniff.php
@@ -4,19 +4,42 @@ namespace SlevomatCodingStandard\Sniffs\TypeHints;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
+use PHPStan\PhpDocParser\Ast\Attribute;
+use PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode;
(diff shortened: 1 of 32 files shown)


kukulich:
@alleknalle Thanks, the PR looks good. However, we need to add new option to enable checks for doc comments.

alleknalle:
> @alleknalle Thanks, the PR looks good. However, we need to add new option to enable checks for doc comments.

@kukulich added an option to enable to sniff for doc comments and updated the docs accordingly.

PR #1851, opened 2026-05-06
FullyQualifiedClassNameInAnnotation parameter ignoredAnnotationNames working for ConstFetchNode

vojmani:
Fixed issue where check for `ignoredAnnotationNames` setting in `FullyQualifiedClassNameInAnnotationSniff` was only working for IdentifierTypeNode, but not ConstFetchNode.

Test assertion decreased by 4 - false positives in `ConstTypeNode` in test data file which were fixed incorrectly while ignored


kukulich:
Thanks.

PR #1855, opened 2026-06-15
Fix (Disallow|Require)TrailingComma*Sniff to handle null parenthesis pointers

HonzaCZ:
**Fix `(Disallow|Require)TrailingComma*Sniff` crash on empty parameter lists (PHP 8.2+)**

## Problem

`DisallowTrailingCommaInDeclarationSniff` (and by extension the other five `*TrailingComma*Sniff` sniffs) crash with a fatal error during a `phpcbf` run when a function has an empty parameter list.

`TokenHelper::findPreviousExcluding()` returns `null` when there are no non-whitespace tokens between the parenthesis opener and closer — i.e. `()`. The sniff then immediately dereferences `$tokens[$pointerBeforeParenthesisCloser]` without a null guard. Using `null` as an array offset is deprecated in PHP 8.1 and fatal in PHP 8.2+ when CodeSniffer's error handler converts `E_DEPRECATED` to a `RuntimeException`.

## Why only `phpcbf`

The crash is triggered indirectly. The following constructor is valid PHP and passes `phpcs` cleanly:

```php
public function __construct(
    // Inject your services here
) {
}
```

When `phpcbf` applies `RequireSingleLineMethodSignature`, it collapses the signature to `public function __construct()`, producing an empty `()`. In the same fix pass, `DisallowTrailingCommaInDeclarationSniff` is invoked on the now-empty parameter list, `findPreviousExcluding()` returns `null`, and the crash occurs.

## Fix

Added a null guard before indexing `$tokens` in all six affected sniffs:

- `DisallowTrailingCommaInCallSniff`
- `DisallowTrailingCommaInClosureUseSniff`
- `DisallowTrailingCommaInDeclarationSniff`
- `RequireTrailingCommaInCallSniff`
- `RequireTrailingCommaInClosureUseSniff`
- `RequireTrailingCommaInDeclarationSniff`

## Reproducer

```php
<?php

class Foo
{
    public function __construct(
        // some comment
    ) {
    }
}
```

Run `phpcbf` with a ruleset that includes both `RequireSingleLineMethodSignature` and `DisallowTrailingCommaInDeclaration` on PHP 8.2+.

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Functions/DisallowTrailingCommaInCallSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Functions/DisallowTrailingCommaInCallSniff.php
@@ -53,7 +53,7 @@ class DisallowTrailingCommaInCallSniff implements Sniff
 		$parenthesisCloserPointer = $tokens[$parenthesisOpenerPointer]['parenthesis_closer'];
 		$pointerBeforeParenthesisCloser = TokenHelper::findPreviousEffective($phpcsFile, $parenthesisCloserPointer - 1);
 
-		if ($tokens[$pointerBeforeParenthesisCloser]['code'] !== T_COMMA) {
+		if ($pointerBeforeParenthesisCloser === null || $tokens[$pointerBeforeParenthesisCloser]['code'] !== T_COMMA) {
 			return;
 		}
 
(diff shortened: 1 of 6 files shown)


kukulich:
Thanks.

Issue #1310, opened 2021-12-04
Enhancement of allowPartialUses


rarila:
It would be good if ``SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.allowPartialUses`` would be enhanced with a list of allowed namespaces (only specified namespaces are allowed to be used partial) and/or a list of namespaces that have to be used partial.

Toflar:
I'm also looking for this feature. Especially because it's best practice for certain packages:

* Doctrine ORM uses `ORM\Entity`, `ORM\Table` etc. in their docs
* Symfony uses it in their validator component and docs (e.g. `#[Assert\Choice(['fiction', 'non-fiction'])]` etc.

Imho using partial references can improve readability a lot depending on the context. For the mentioned packages above for example, you combine those in the same class quite often and using partial references makes it super easy to differentiate them quickly from each other.

Is there a way to financially support implementing this feature?

Toflar:
Just as an info to anybody potentially having subscribed to this issue. I gave this a try in https://github.com/slevomat/coding-standard/pull/1856 😎

Toflar:
@kukulich this can be closed now that https://github.com/slevomat/coding-standard/pull/1856 has been released 😊

PR #1856, opened 2026-06-18
Add support to allow partial use rules per namespace

Toflar:
Implements feature request in https://github.com/slevomat/coding-standard/issues/1310

It adds namespace-aware control to `SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly` partial-use handling.

As already explained in the original issue, the reason why I would like to have this is that some codebases intentionally use partial references for selected packages, especially Doctrine ORM and Symfony Validator.

You can now either just allow them:

```xml
<rule ref="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly">
	<properties>
		<property name="allowPartialUses" value="false"/>
		<property name="namespacesAllowedToUsePartially" type="array">
			<element value="Doctrine\ORM\Mapping as ORM"/>
			<element value="Symfony\Component\Validator\Constraints as Assert"/>
		</property>
	</properties>
</rule>
```

or even enforce them:

```xml
<rule ref="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly">
	<properties>
		<property name="allowPartialUses" value="false"/>
		<property name="namespacesRequiredToUsePartially" type="array">
			<element value="Doctrine\ORM\Mapping as ORM"/>
			<element value="Symfony\Component\Validator\Constraints as Assert"/>
		</property>
	</properties>
</rule>
```

### DIFF
--- a/SlevomatCodingStandard/Helpers/NamespaceHelper.php
+++ b/SlevomatCodingStandard/Helpers/NamespaceHelper.php
@@ -13,10 +13,12 @@ use function explode;
 use function implode;
 use function ltrim;
 use function min;
+use function preg_match;
 use function sprintf;
 use function strcasecmp;
 use function strcmp;
 use function strpos;
+use function trim;
 use const T_NAME_FULLY_QUALIFIED;
 use const T_NAMESPACE;
 use const T_NS_SEPARATOR;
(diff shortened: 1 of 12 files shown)


Toflar:
Updated the PR so it now also allows namespace aliases properly supporting the `Symfony\Component\Validator\Constraints as Assert` as well. Initial PR description updated 😎

kukulich:
@Toflar Looks good. Can you please squash it to one commit?

Toflar:
Done 😊

PR #1859, opened 2026-06-28
Fix ReadonlyClassSniff to not to require class marked as readonly whe…

kamil-zacek:
…n it is not final, not marked as AllowDynamicProperties and without traits

### DIFF
--- a/SlevomatCodingStandard/Sniffs/Classes/ReadonlyClassSniff.php
+++ b/SlevomatCodingStandard/Sniffs/Classes/ReadonlyClassSniff.php
@@ -4,6 +4,7 @@ namespace SlevomatCodingStandard\Sniffs\Classes;
 
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
+use SlevomatCodingStandard\Helpers\AttributeHelper;
 use SlevomatCodingStandard\Helpers\ClassHelper;
 use SlevomatCodingStandard\Helpers\FunctionHelper;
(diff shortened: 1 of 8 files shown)


doc/classes.md
```diff
@@ -234,7 +234,14 @@ Sniff provides the following settings:
 
 #### SlevomatCodingStandard.Classes.ReadonlyClass 🔧
 
-Reports classes where all promoted constructor properties are declared as `readonly` and suggests marking the whole class as `readonly`.
+Reports classes where all promoted constructor properties and class body properties are declared as `readonly` and suggests marking the whole class as `readonly`.
+
+Suggestion is reported only for classes that are `final`, do not `extends` another class, do not use traits, and do not have `#[\AllowDynamicProperties]`.
```
kukulich: `that are `final`` - it's not true anymore.

kukulich: `do not use traits` - same

SlevomatCodingStandard/Sniffs/Classes/ReadonlyClassSniff.php
```diff
@@ -125,7 +130,19 @@ public function process(File $phpcsFile, int $classPointer): void
 			}
 		}
 
-		if ($this->classExtendsAnotherClass($phpcsFile, $classPointer)) {
+		$isFinal = ClassHelper::isFinal($phpcsFile, $classPointer);
+		if (
```
kukulich: Can you split it to more `if` ?

kukulich (changes requested):

